Issue #7, opened 2018-01-29
Should a request sent from client to server include browser cookies?


blair55:
I'm observing that a Fable.Remoting RPC POST request to the (Suave) server does not include the Cookie header, so domain cookie values are not sent. This makes session management difficult.

I am wondering is this by design? Or perhaps an XMLHttpRequest restriction? I had thought you might not have to cater for this concept in the client lib at all, as the cookie header would be included automatically by the browser...?

What do you suggest?

Zaid-Ajaj:
Hello @blair55, 
>I'm observing that a Fable.Remoting RPC POST request to the (Suave) server does not include the Cookie header, so domain cookie values are not sent. This makes session management difficult.

Indeed, this library only uses the request body data for communication. 

> I am wondering is this by design?

Not really, although it is recommended you use RPC for user authentication/session management with stateless tokens, there should be no restriction to use other machanisms like cookies assuming you are using this library as part of a bigger application. 

> as the cookie header would be included automatically by the browser...?

AFAIK the cookies should be sent along with the POST request automatically, but maybe since I am explicitly only using the body request, the cookies are ignored. 

>What do you suggest?

I will try to add the cookie header manually

blair55:
Amazing, what a great response! I'm glad you could verify my thoughts.

Do you have an example of using stateless tokens with Fable.Remoting? I think I will stick with cookies if you can make it available, but would be good to appreciate the alternative.

Great library btw!
Thanks

Zaid-Ajaj:
Thank you for your kind words :smile: 

I now have published a new version (1.4.0) of the Fable.Remoting.Client library where the browser cookies are included in the RPC request using the `Cookie` header with value `document.cookie`. I verified that it works on my integration tests project:

![screenshot from 2018-01-30 21-35-56](https://user-images.githubusercontent.com/13316248/35590768-18717a18-0608-11e8-8093-ea05ec4bb9ee.png)

Can you check that it works for you as well? 

> Do you have an example of using stateless tokens with Fable.Remoting?

Unfortunately, as of now, I do not have such example :( but I will try to put on together really soon when the time permits

blair55:
Perfect! That's having the effect of sending the document cookie as expected.

Thanks for closing so quickly. Great work!

Zaid-Ajaj:
Great, then this is resolved! I will close. Just let me know if you come across any issue :smile:

PR #8, opened 2018-02-07
Support an optional authorization header

Nhowka:
Enable usage of a token or other form of authentication in the Authorization header when creating the proxy.

### DIFF
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -117,8 +117,8 @@ module Proxy =
         )
         |> List.ofSeq
 
-    /// Creates a secure proxy using a custom endpoint and a route builder
-    let [<PassGenerics>] createSecureWithEndpointAndBuilder<'t> (endpoint: string option) (routeBuilder : string -> string -> string) (auth: string option) : 't = 
+    
+    let [<PassGenerics>] private createSecureWithEndpointAndBuilderImpl<'t> (endpoint: string option) (routeBuilder : string -> string -> string) (auth: string option) : 't = 
         // create an empty object literal
         let proxy = obj()
         let typeName = typeof<'t>.Name
@@ -132,35 +132,39 @@ module Proxy =
             let fieldName = fst field
             setProp fieldName (proxyFetch typeName fieldName returnType endpoint routeBuilder auth) proxy
         )
-        unbox proxy
+        unbox proxy    
     /// Creates a proxy using a custom endpoint and a route builder
     let [<PassGenerics>] createWithEndpointAndBuilder<'t> (endpoint: string option) (routeBuilder : string -> string -> string): 't = 
-        createSecureWithEndpointAndBuilder<'t> endpoint routeBuilder None
+        createSecureWithEndpointAndBuilderImpl<'t> endpoint routeBuilder None
 
     /// Creates a proxy that routes method calls to /typeName/methodName
     let [<PassGenerics>] create<'t>  : 't = 
-        createSecureWithEndpointAndBuilder<'t> (Some "/") (sprintf "/%s/%s") None
+        createSecureWithEndpointAndBuilderImpl<'t> (Some "/") (sprintf "/%s/%s") None
 
     /// Creates a proxy using a custom endpoint and the default route builder.
     [<PassGenerics>]
     let createWithEndpoint<'t> (endpoint: string) : 't = 
-        createSecureWithEndpointAndBuilder<'t> (Some endpoint) (sprintf "/%s/%s") None
+        createSecureWithEndpointAndBuilderImpl<'t> (Some endpoint) (sprintf "/%s/%s") None
 
     /// Creates a proxy using the default endpoint = "/" and a custom route builder
     [<PassGenerics>]
     let createWithBuilder<'t> (routeBuilder: string -> string -> string) : 't = 
-        createSecureWithEndpointAndBuilder<'t> None routeBuilder None
+        createSecureWithEndpointAndBuilderImpl<'t> None routeBuilder None
+    
+    /// Creates a secure proxy using a custom endpoint and a route builder
+    let [<PassGenerics>] createSecureWithEndpointAndBuilder<'t> (endpoint: string option) (routeBuilder : string -> string -> string) (auth: string): 't = 
+        createSecureWithEndpointAndBuilderImpl<'t> endpoint routeBuilder (Some auth)
 
     /// Creates a secure proxy that routes method calls to /typeName/methodName
-    let [<PassGenerics>] createSecure<'t>  : string option -> 't = 
-        createSecureWithEndpointAndBuilder<'t> (Some "/") (sprintf "/%s/%s")
+    let [<PassGenerics>] createSecure<'t>  (auth: string): 't = 
+        createSecureWithEndpointAndBuilderImpl<'t> (Some "/") (sprintf "/%s/%s") (Some auth)
 
     /// Creates a secure proxy using a custom endpoint and the default route builder.
     [<PassGenerics>]
-    let createSecureWithEndpoint<'t> (endpoint: string) : string option -> 't = 
-        createSecureWithEndpointAndBuilder<'t> (Some endpoint) (sprintf "/%s/%s")
+    let createSecureWithEndpoint<'t> (endpoint: string) (auth: string) : 't = 
+        createSecureWithEndpointAndBuilderImpl<'t> (Some endpoint) (sprintf "/%s/%s") (Some auth)
 
     /// Creates a secure proxy using the default endpoint = "/" and a custom route builder
     [<PassGenerics>]
-    let createSecureWithBuilder<'t> (routeBuilder: string -> string -> string) : string option -> 't = 
-        createSecureWithEndpointAndBuilder<'t> None routeBuilder
+    let createSecureWithBuilder<'t> (routeBuilder: string -> string -> string) (auth: string) :  't = 
+        createSecureWithEndpointAndBuilderImpl<'t> None routeBuilder (Some auth)
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -33,7 +33,7 @@ module Proxy =
     let private jsonParse (content: string) : obj = jsNative
     [<Emit("JSON.stringify($0)")>]
     let private stringify (x: obj) : string = jsNative
-    let private proxyFetch typeName methodName returnType (endpoint: string option) (routeBuilder: string -> string -> string) =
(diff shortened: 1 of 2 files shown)


Zaid-Ajaj:
Hello @Nhowka this is very interesting, don't mind the travis-ci checks, it happens sometimes. I will have to run the integration tests locally (sometime this evening) to verify it working as expected, then I will merge.

Thank you this!

Nhowka:
Looks like there was a case where Suave tried to bind onto an already bound port. Is it ok to change the tester from random to a sequential one?

Nhowka:
I really underestimated the work of having a sequential port generation...

Zaid-Ajaj:
> Looks like there was a case where Suave tried to bind onto an already bound port. 

Yeah, random ports are quite a hack to be honest :sweat_smile: the atomic increment seemed to have solved the issue :heart: thank you!

Zaid-Ajaj:
Just published the next version of the Fable.Remoting.Client (v1.5.0) should be available on Nuget soon 
 after nuget finishes validation and indexing :smile:

Nhowka:
Thank you! Now it should be easier using the proxy with Saturn without changing the pipeline for authentication, and maybe on the SAFE-BookStore too.

blair55:
Hi both. I'm trying to understand the change here... whats support is there in the adapter to read this Auth header?

Say I use `Proxy.createSecureWithBuilder` and provide the bearer token retrieved on startup, how do I access the value of the header server-side? The protocol implementation is designed to bind only the body of the request to the function arg. I see there is the `FableGiraffeAdapter.handleRequest` (or similar) which does have access to the HttpContext and therefore the headers, but I don't think that is designed to be used in the same way.

Could you clarify how you expect this feature to be used? :-)

Nhowka:
You don't use it inside the adapter but before it. e.g. In Giraffe you have a `requiresAuthentication` that you can put it before you pipe the adapter into.

Adapting the [Giraffe example](https://github.com/giraffe-fsharp/Giraffe/blob/master/DOCUMENTATION.md#requiresauthentication):
```fsharp
let notLoggedIn =
    RequestErrors.UNAUTHORIZED
        "Basic"
        "Some Realm"
        "You must be logged in."

let mustBeLoggedIn = requiresAuthentication notLoggedIn

let handler = FableGiraffeAdapter.httpHandlerFor yourImplementation

let webApp =
    choose [
        route "/"     >=> text "Hello World"
        route "/user" >=>
            mustBeLoggedIn >=> handler
    ]
```

Zaid-Ajaj:
@blair55 It is true that you can't access the headers in the automatic request handler of Fable.Remoting but you can still do that *in an earlier point* of the request pipeline. i.e.:
```fs
// example in Suave
let webApp = requireAuthenticated >=> fableWebPart
```
Where `requireAuthenticated` is a handler that checks whether or not the request has valid authentication headers before propagating the request to the next handler (`fableWebPart` in this case)

Maybe @Nhowka can provide a more concrete example :smile:

blair55:
Thanks both, I understand that you can 'catch' the unauthorized request before entering into the remoting handler, but that raises two further questions.

1 - there's no client side support to handle a response with unauthorized status:
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Proxy.fs#L96
By providing the Auth header value when creating the proxy, shouldn't you also be able to express what happens if the value is considered invalid?

2 - it would be useful to access the user token in application code so that I can customise the response. e.g. get settings for user. If the identity is already present in the header I'd like to be able to use that value in deciding what the response should be instead of having to carry an extra parameter in the request object. (sorry, not really a question!)

Nhowka:
1. Valid point. Probably we can do an extra match with `401 when auth.IsSome` with a different `onAuthError` handler.

1. Probably you can do some sort of partial application to have the user on context without breaking the `'a -> Async<'b>` expectation. I don't disagree with you that this isn't the best choice, but I don't know what we should do to keep it simple and generic enough.

blair55:
Thanks for addressing point 1. As for 2, I'd agree partial application could be a nice fit. Or maybe a tuple for the first argument `(Context, 'a) -> Async<'b>`. Would like to know what @Zaid-Ajaj thinks.

Zaid-Ajaj:
@blair55 Point 1 makes a lot of sense, and @Nhowka addressed it in his [pull request](https://github.com/Zaid-Ajaj/Fable.Remoting/pull/13) 

However, point 2, to me at least, breaks the abstraction this library provides. When I wrote the library, I didn't only want to automate data serialization between client/server but *also* forget all about http and just think of your application in terms of simple stateless functions you call from the client. 

You might ask, what does "forgetting Http" mean in the context of authenticated users? Well, it means that you should model authentication in terms of simple stateless functions, the following types/protocol describes a situation with user authentication:
```fs
// Shared.fs

type Token = string

type LoginInfo = { Username: string; Password: string }

// possible errors when logging in
type LoginError = 
    | UserDoesNotExist
    | PasswordIncorrect
    | AccountBanned

// a request with a token
type SecureRequest<'t> = { Token : string; Content : 't }

// authentication/authorization errors     
type AuthError = 
   | UserTokenExpired
   | TokenInvalid
   | UserDoesNotHaveAccess

// some domain model
type Book = { Id: int; Title: string; (* other propeties *) }

// the actual protocol
type IBookStoreApi = {
    // login to acquire an auth token   
    login : LoginInfo -> Async<Result<Token, LoginError>>
    // "public" function: no auth needed
    searchBooksByTitle : string -> Async<list<Book>> 
    // secure function, requires a token
    booksOnWishlist : Token -> Async<Result<list<Book>, AuthError>>, 
    // secure function, requires a token and a book id
    removeBookFromWishlist : SecureRequest<int> -> Async<Result<bool, AuthError>>
    // etc . . . 
}
```
As you can see, by only looking at `IBookStoreApi` you can understand what the whole application does. Introducing contextual information makes it a leaky abstraction and someone would have to track the state coming from the request pipeline before and after the remoting handler, it is still fine if you do introduce cookies/auth headers etc if you find it easier that way. Personally, I think it just complicates the model.

blair55:
Ok I understand your point. I can see the value passing the token explicitly as part of the function and not by contextual means. And thank you for the detailed example.

But now I am conflicted by the options the library provides.

By offering `Proxy.createSecureWithBuilder` you have not forgotten about http. You have implemented half the http story so it leads down a dead end. I can send a header but I cannot read it?!

How does @Nhowka's example above fit with your AuthError type?

```
type AuthError = 
   | UserTokenExpired
   | TokenInvalid
   | UserDoesNotHaveAccess

let webApp =
    choose [
        route "/"     >=> text "Hello World"
        route "/user" >=>
            mustBeLoggedIn >=> handler
    ]
```

> Personally, I think it just complicates the model.

I agree. As someone who is accustomed to using http frameworks, I need an RPC framework to be more opinionated and offer clearer guidance. I don't feel that I am falling into the pit of success because these options do not enforce the stateless functions opinion.

Nhowka:
I just wanted to use the proxy with a framework where I could protect an endpoint in under 3 lines. I agree with you about it being incomplete as the client and the server are not sharing the same protocol. I'll think in something on the server with more control about what can happen when you have a rejected header or at least give a way to read it.

blair55:
@Nhowka I appreciate your intention - I opened #7 with the same goal! But IMO this isn't a question of mechanics now, its a question of principles. What identity do you want this framework to have?

Nhowka:
To be honest, the main reason I like this project is that it enables me to just write a record, an implementation and have a complete API that I can call as if native. When I couldn't use it, the main timesink I had was debugging the promises and make the server understand what the client was asking for, as I had no power on the server-side serialization.

I wouldn't use some framework if it just promised me that I would end with a better model. I want to use this because I saves my time.

As I can leverage the authorization features on the server, protecting some endpoints is something I can do very easily by just passing a header. That for me is almost magical and priceless. It's something that took me hours to get it right in other project and now with this it will take less than a minute in the next.

So I think I'm doing that for my people, the lazy.

Zaid-Ajaj:
@blair55 
> You have implemented half the http story so it leads down a dead end. I can send a header but I cannot read it?!

> How does @Nhowka's example above fit with your AuthError type?


Good question, and there is still a clean solution to this: **dependent handlers!** You can read the header from **inside** the remoting handler without compromising the shape of the API. @Nhowka This is could be interesting for you as well. I wrote a full sample to demonstrate:

Given the following protocol:
```fs
type Token = string 

type AuthError = 
   | UserTokenExpired
   | TokenInvalid
   | UserDoesNotHaveAccess

type IServerApi = {
    // No token required in the input of the function
    getBooks : unit -> Async<Result<string[], AuthError>>
}
```
Notice how the function `getBooks` does not have a token in the input parameter. 

Now, your implementation of `IServerApi` will be **dependent** on the authorization token provided from the request pipeline:

```fs
// Verify that the authorization token is valid
let validateToken token = token <> ""

// Token as a dependency for the server implementation
let server token = {
    getBooks = fun () -> async {
        let result =
          match validateToken token with 
          | false -> Error TokenInvalid
          | _ -> Ok [| "Harry Potter" |]
        
        return result
    }
}
```
Create the webpart from a token:
```fs
// fableWebPart : Token -> WebPart
let fableWebPart token = 
    FableSuaveAdapter.webPartFor (server token)

/// Read authorization header from request and pass it to the remoting handler
let authorizedFableWebpart (req: HttpRequest) = 
    req.headers
    |> List.tryFind (fun (key, _) -> key = "authorization" )
    |> Option.map snd
    |> function 
        | Some authToken -> fableWebPart authToken
        | None -> fableWebPart ""

let webApp = request authorizedFableWebpart

startWebServer defaultConfig webApp
```
Now test the API without sending an authorization header:
screenshot from 2018-02-16 02-00-16

But now when you do send an authorization token, you get the correct response:

screenshot from 2018-02-16 02-03-08

Here is a [gist](https://gist.github.com/Zaid-Ajaj/fd475896f7057f148cd06135d4d8a1e6) of the whole app.

Zaid-Ajaj:
> But now I am conflicted by the options the library provides.

I hope this is no longer the case. I think this way the library makes everyone happy: 
 - Users who want a clean, strict API model to work against = forget http
 - Users who want to leverage the automatic type convertions as part of a bigger application

I am eager to hear your feedback!

blair55:
@Zaid-Ajaj, thanks again for the full example. I did not know about the dependent handler approach, I withdraw my comment about it being a dead end! I will use this option and let you how I get on.

I also recognise that the flexibility afforded by the two approaches might make the library an option in more use cases, as opposed to just diluting its purpose.

@Nhowka thanks also for your response, I too am one of the lazy! :-D

Zaid-Ajaj:
@blair55 I am happy to hear that. I will be working on a proper documentation website soon, so everything will be there for reference.

PR #9, opened 2018-02-08
Saturn

Nhowka:
Adapter for use with [Saturn](https://github.com/SaturnFramework/Saturn).


Zaid-Ajaj:
Wow, this is awesome! :heart: the code looks good, I will merge right away. We still need to add the integration tests though, but that can be done later.

Zaid-Ajaj:
The code can be simplified by referencing the already implemented methods from Fable.Remoting.Giraffe. I will take a look

Nhowka:
Good idea! The code barely changed so that should be easier to maintain!

Krzysztof-Cieslak:
Oh wow ❤️

PR #11, opened 2018-02-12
paket.restore.cached should never be commited

forki:
please gitignore /paket-files


Zaid-Ajaj:
I will, thank you :pray:

PR #12, opened 2018-02-12
Temporary fix for Saturn

Nhowka:
Use the previous Giraffe version until the next Saturn version with Giraffe 1.0.0


Zaid-Ajaj:
Awesome! I was going to wait for Saturn to patch Giraffe to v1.0 but I guess we might as well start with this version

Zaid-Ajaj:
@Nhowka Published Fable.Remoting.Saturn 1.0 should be available on Nuget soon. Thanks a lot for this :heart:

Nhowka:
I'm worried about the dependencies as Saturn has a dependency on Giraffe being at least `0.1.0-beta-700`. It will load the version 1.0.0 instead. Could we have the Giraffe version pinned to the same until the next Saturn release?

Zaid-Ajaj:
I don't think that is necessary, I *think* it will load `Giraffe 1.0` only when both `Saturn` AND `Giraffe 1.0` are being used at the same time, which highly unlikely, see [this issue](https://github.com/fsprojects/Paket/issues/3039) (unless you have a mono-repo like this repo :smile:) .

Nhowka:
I did a `paket install` on a project and stuff failed to build until I pinned Giraffe to `0.1.0-beta-700`. Put only `Saturn` and `Fable.Remoting.Saturn` as dependencies.

Zaid-Ajaj:
I guess it is okay for now to pin Giraffe to `0.1.0-beta-700` because I already published Fable.Remoting.Giraffe that depends on `Giraffe 1.0`, so temporarily that would be find util Saturn is patched

PR #13, opened 2018-02-13
Support custom handlers on unauthorized and forbidden

Nhowka:
Address a concern raised on #8 discussion. That way the client can do something different when the failure is from an unauthorized or forbidden response.

### DIFF
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -15,11 +15,21 @@ module Proxy =
         response: Response
     }
 
+    let mutable private authHandler : Option<string option -> unit> = None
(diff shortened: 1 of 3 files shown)


Nhowka:
Pinging @blair55 for feedback.

blair55:
Yep, makes sense. I question the scalability of the design. i.e. could we end up with a mutable  placeholder function for every http status code/range? Otherwise, a worthwhile improvement in my book. 👍

Nhowka:
We could have a map with the relation `error -> handler`, but other errors shouldn't happen, at least in normal usage.

Zaid-Ajaj:
Published Fable.Remoting.Client 1.6 that includes this feature, I am very happy with your feedback and input on this library. Thanks a lot :pray: :pray: :heart:

PR #10, opened 2018-02-08
Simplify saturn code reusing most of giraffe

Nhowka:

PR #14, opened 2018-02-20
Redo saturn simplification

Nhowka:
Now that Saturn was updated, we can redo #10


Zaid-Ajaj:
Nice job, you are quite up to date :smile: thanks!

Zaid-Ajaj:
Fable.Remoting.Saturn 1.1 published with latest deps. Should be available in few hours

PR #15, opened 2018-03-01
Multiple arguments proof-of-concept

Nhowka:
I changed some stuff to support multiple arguments. Shouldn't be a breaking change, but should be discussed more.


Zaid-Ajaj:
Nice addition to the library, usually this isn't really needed because you can use tuples or records with fields as the parameters

Zaid-Ajaj:
I will need to test it locally with the integration tests before publishing new packages :smile:

Nhowka:
I had a problem with Elmish when using it with the `Cmd.ofAsync` because of the partial applying, but when removing the `arg` and passing directly an `Async<_>` it worked well. I'm doing a Saturn inspired API to remove even further duplication and make the usage of the logs and handlers friendlier. You can check it [on my fork](https://github.com/Nhowka/Fable.Remoting/tree/CEBasedAPI).

Zaid-Ajaj:
The CE looks nice, I don't mind having it in the library. Personally, I prefer not using CE if it only adds syntactic sugar because it just adds even more to the magic the library already has. But again, it makes a nice match with Saturn's api :smile:

Zaid-Ajaj:
I sent you an invitation to become a collaborator on this repo :smile: Sorry for not having enough time to keep up with your awesome work

Nhowka:
Thank you for your kind words! I've changed the readme to reflect the changes on the API. It made setting logs and handlers more explicit at least, good for cases when you want different stuff to happen when having errors on different levels without depending on the global variable.

Zaid-Ajaj:
@Nhowka I was able to run the integration tests with your latest changes. Everything works as it should :heart: All packages were updated

Nhowka:
I'm happy about not breaking anything, but there is a bug in the multiple arguments when they are partially applied. Discussed it a little on fable-elmish/elmish#142.

I'm working on a real fix.

PR #21, opened 2018-03-08
Ce based api

Nhowka:


Zaid-Ajaj:
Awesome! I will update the packages as soon as I run the integration tests and also add some more tests for muti-arg functions.

Nhowka:
After today's discussion, I'm thinking about turning the CE into an abstract class and making `Run` abstract, so we don't need to open `Fable.Remoting.Server`, just implement the base class.

PR #24, opened 2018-03-13
General customizations

Nhowka:
With these changes there is support for sending custom headers, overriding the server response and status code, overriding the client's handler so it can interpret the overridden response. Also, there is support for aborting the request and let it flow through the pipeline, making it possible to create different versions on the same endpoint by requesting it with a different header.

Needs more documentation, but has some integration tests.


Nhowka:
It's probably better to give access to the real context instead of forcing it to fit into a common interface. These features are target to advanced users that want more power, so they should be trusted with it! 😋

Zaid-Ajaj:
Looks very interesting, indeed giving access to the original context makes it much easier to work with. I would like to see some usage samples of these, to see how the code would look like for end users :smile:

Nhowka:
In this latest commit, I created an endpoint that gives a different response based on the request header:

```fs
type IVersionTestServer = {
    v4 : unit -> Async<string>
    v3 : unit -> Async<string>
    v2 : unit -> Async<string>
    v1 : unit -> Async<string>
}

let versionTestBuilder _ _ = "/api/version/data"
```

With implementation

```fs
let versionTestServer : IVersionTestServer = {
    v4 = fun () -> async {return "v4"}
    v3 = fun () -> async {return "v3"}
    v2 = fun () -> async {return "v2"}
    v1 = fun () -> async {return "v1"}
}
```

On the server i created the webpart with
```fs
let isVersion v (ctx:HttpContext) =
  if ctx.request.headers |> List.contains ("version",v) then
    None
  else
    Some {ResponseOverride.Default with Abort = true}

let versionTestWebPart =
  remoting versionTestServer {
    use_logger (printfn "%s")
    with_builder versionTestBuilder
    use_custom_handler_for "v4" (isVersion "4")
    use_custom_handler_for "v3" (isVersion "3")
    use_custom_handler_for "v2" (isVersion "2")
  }
```

and on the client

```fs
let versionTestServer = Proxy.remoting<IVersionTestServer> {
    with_builder versionTestBuilder
    add_custom_header_for "v2" ("version",2)
    add_custom_header_for "v3" ("version",3)
    add_custom_header_for "v4" ("version",4)
    }
```

This way you can create different requests with different responses on the same endpoint, updating new clients without breaking the old ones. It's not something I expect to be common, but now it's possible.

pavelm:
How would you handle the case when a function needs to access the currently logged in user? That is a common approach for user facing apps in that they pass around the authorization token and are implicitly "logged in".

Nhowka:
I'm not sure about Suave, but in the Giraffe's HttpContext you have a `User` property that you can check if it is logged or not. But you can just pass a token and create a proxy that will send the auth header and expect it to only be reached if the user was validated up in the pipeline.

Now you have the option to handle it yourself, but I would recommend to just have the Unauthorized and Forbidden handlers up so you can do something if the user is rejected.

pavelm:
That makes sense, I'm just confused how to handle it from the server side. 

For example
`let serverApi : IServerApi = {
    forUser = fun user -> async {return "user logged in: " + user.Name }
}`

or can you specify you need access to the RequestContext
`let serverApi : IServerApi = {
    forUser = fun ctx -> async {return "user logged in: " + ctx.User.Name }
}`

Nhowka:
You wouldn't have access of the context as an argument.

```fs
let webPart = remoting serverApi {
    use_custom_handler_for "forUser"
        (fun ctx ->
            Some ResponseOverride.Default.withBody("user logged in: " + ctx.User.Name))
    }
```

When you access the context with the custom handler, you have the option to override the response that the original function would give. That way it is an entirely optional feature that can give power to do lots of things without getting in the way on the signature of the function.

pavelm:
But sometimes the function needs the context or just the user as an input. For example to load the user's profile, or their favorite items, etc. 

It's not just about overriding the output, but allowing the context or user as input to the server function. 

I know it violates getting in the way of the signature, but I think it's also a viable use case.

Nhowka:
But how would you call it from the client? That's the problem we couldn't solve yet...

Zaid-Ajaj:
@Nhowka This makes a lot of scenario's possible but I don't think we should recommend users to go this way when they want to integrate `Fable.Remoting` into a bigger application, it would be only for rare use-cases like the one you gave an example of, otherwise the user code would clutter up real quick if they abuse this feature and I see it happening for people who want to "just get it working".

Zaid-Ajaj:
@pavelm We have discussed this before, a lot :smile: The recommendation was always to [model security features](https://zaid-ajaj.github.io/Fable.Remoting/src/modeling-authentication.html) in terms of `Fable.Remoting` to get cleanest solution. 

Another solution which is about to get in the documentation is to use *dependent* handlers, these are remoting handlers that can use the request context without sacrificing the shape the remoting api while still being a clean solution, see [this detailed comment with full example](https://github.com/Zaid-Ajaj/Fable.Remoting/pull/8#issuecomment-366115586) demonstrating this concept. 

As for your use case, if you want a user to be the input of your function, just send the id or username or auth token as the input of the function and then read the user information from there, if you have auth implemented with cookies, you can read the cookie data from the client and send it off to the server.

Nhowka:
@Zaid-Ajaj I chose a lengthy name for discourage abusing it. Maybe it will never be really used, as using it makes the normal usage seems cleaner and complete enough.

But now people that want do weird stuff have a way to do it. 😙

Nhowka:
Maybe these features should be in a separated place in documentation, like "Advanced features" or something like that.

Zaid-Ajaj:
@Nhowka There is already an "advanced" section in the [docs](https://zaid-ajaj.github.io/Fable.Remoting/) for any non-default feature. An "awkwardly advanced" section sounds more appropriate :joy:

Nhowka:
@Zaid-Ajaj Hahahaha! Maybe "Power-User features" could be an appropriate name for inappropriate usage! 😂

Nhowka:
There! With this last commit now we can do:
```fs
type IServer<'ctx> = {
   callWithCtx: 'ctx -> Async<string>
}
```
Create the server: 
```fs
let contextImpl = {
  callWithCtx = fun (ctx:HttpContext) -> async{ return ctx.request.path }
  }

let contextTestWebApp =
    remoting contextImpl {
        use_logger (printfn "%s")
        with_builder routeBuilder
    }
```
Create the client:
```fs
let contextTestServer = Proxy.remoting<IContextTest<unit>>{
    with_builder routeBuilder
}
```

And then call it as `contextTestServer.callWithCtx ()` and have the context properly injected.

Nhowka:
That was the last feature that I wanted to add. Now there is no need to have a `unit -> Async<_>`, you can have just an `Async<_>`.

Zaid-Ajaj:
Wow @Nhowka so many cool features, this is seriously awesome, especially reading the request context! I was writing the docs on accessing the request context and now I am seeing this :heart: :heart: :heart:

Can't wait for this to get merged. I will be tweeting about this when we are done with the features and docs, do you have a twitter account?

Zaid-Ajaj:
@Nhowka Can you please review the docs to make sure that this is how meant the feature to be used? 

See [Accessing Request Context](https://zaid-ajaj.github.io/Fable.Remoting/src/request-context.html)

Nhowka:
Thanks! I was postponing merging until I had these features documented, but I just saw you did an excelent job! ❤

I'll leave the "Power-User features" for later, but I enabled the stuff that handles the tooltip help as seen here: https://github.com/SaturnFramework/Saturn/pull/36. If everything is green, it's ready for merging!

Nhowka:
My twitter is kinda abandoned since 2015, but it's [DiEsmerio](https://twitter.com/DiEsmerio) 😅

Zaid-Ajaj:
> I'll leave the "Power-User features" for later 

I will document those later, after we publish the current features and recieve some feedback about them.

Thanks for merging :tada:, I will be able to do some more testing very soon, I also need to test the things I documented to make sure the library does what it says it does :smile:  only after that, I will be able to publish the new feature (sorry if it will take some time on end of things, I hope you don't mind that) 

> My twitter is kinda abandoned since 2015

It would be cool if you started using it again, devs of F# community usually share the cool things they make on twitter and it is mostly really interesting stuff (like this library :wink:)

Nhowka:
No problem! It's better to publish after everything is well tested than needing to publish just bug-fixes. We should probably normalize the custom operations, mainly on the client.

Server:
- `with_builder`
- `use_logger`
- `use_error_handler`
- `use_custom_handler_for`

Client:
- `at_endpoint`
- `use_error_handler`
- `use_auth_error_handler`
- `use_forbidden_error_handler`
- `with_token`
- `with_builder`
- `add_custom_header`
- `add_fable_header`
- `add_custom_header_for`
- `add_fable_header_for`
- `use_custom_handler_for`

I'm guessing using `use` as prefix would make sense for every method, making intellisense more useful too.

I'll try to give a second chance on twitter! 😗

Zaid-Ajaj:
@Nhowka So after adding a couple of tests I got a little bit impatient and published the packages and updated the docs because everything seems to work like it should and I can't wait to spread the word :smile: :smile: 

I also added an alias for `with_builder` called `use_route_builder` which I think is a bit more specific than `with_builder` (we will use the `use_` prefix anyway)

Nhowka:
The alias is a great idea! It would be nice if `CustomOperation` were allowed multiple times, then it would be easier to make them without needing to add a new member, not sure why it isn't allowed...

I'm worried about the reflection features in Fable 2. I hope we can continue to provide such a cool library when it's released.

Zaid-Ajaj:
Yeah Fable 2 won't support Reflection at the early stages, hopefully later on it will be supported on a stable release. Otherwise, I was thinking on writing a Fable compiler plugin that dumps the metadata of types somewhere so that it is still readable from Client, this however means that users must install and reference the plugin in the Fable options. Hopefully we won't have to resort to that solution

Nhowka:
Another thing, how does the nuget dependencies work? I saw that it's still having a `Microsoft.CSharp`. Shouldn't removing the dependency on `Fable.Remoting.Reflection` also remove this one? 🤔

Zaid-Ajaj:
Yeah it should be removed, I will take a look

Zaid-Ajaj:
It seems that even if you don't use it but still have it in `paket.references`, it will be referenced as a dependency, man, I guess I need to re-publish yet again...

Nhowka:
I removed from the dependencies and totally forgot references... Sometimes paket is too smart 😅

Zaid-Ajaj:
That's fine though, packages updated :smile:

PR #30, opened 2018-03-26
[WIP] Fix Union with DateTime JSON conversion 

Zaid-Ajaj:
This type:
```fs
type UnionWithDateTime = 
    | Date of DateTime
    | Int of int
```
fails this test:
```fs
testCase "Union with DateTime conversion" <| fun () ->  
    let dateInput = DateTime.Now
    let serialized = serialize (UnionWithDateTime.Date dateInput) 
    let deserialized = deserialize<UnionWithDateTime> serialized 
    match deserialized with 
    | Int _ -> fail() 
    | Date dateOutput ->
        Expect.equal dateInput.Second dateOutput.Second "Seconds are the same"
        Expect.equal dateInput.Minute dateOutput.Minute "Minutes are the same" 
        Expect.equal dateInput.Hour dateOutput.Hour "Hours are the same" 
        Expect.equal dateInput.Day dateOutput.Day "Days are the same" 
        Expect.equal dateInput.Month dateOutput.Month "Months are the same" 
        Expect.equal dateInput.Year dateOutput.Year "Year are the same" 
        Expect.equal dateInput.Kind dateOutput.Kind "Kinds are the same"
```
with message:
```fs
Converter Tests/Union with DateTime conversion failed in 00:00:00.7470000.
Hours are the same. Actual value was 23 but had expected it to be 21.
```

### DIFF
--- a/Fable.Remoting.Json.Tests/FableConverterTests.fs
+++ b/Fable.Remoting.Json.Tests/FableConverterTests.fs
@@ -37,7 +37,7 @@ let converterTest =
             | otherwise -> fail()
 
         testCase "Union with DateTime conversion" <| fun () ->  
-            let dateInput = DateTime.Now
+            let dateInput = DateTime.Now.ToUniversalTime()
             let serialized = serialize (UnionWithDateTime.Date dateInput) 
             let deserialized = deserialize<UnionWithDateTime> serialized 
             match deserialized with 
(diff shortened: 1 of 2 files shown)


Nhowka:
On Travis the error was
`Kinds are the same. Actual value was Local but had expected it to be Utc.`

That should explain the error and hour difference.

Nhowka:
`DateTime.UtcNow` passes

Zaid-Ajaj:
I see now, it seems that every `DateTime` is converter to UTC by the JSON converter as this what Fable's `ofJson` expects. So testing that the output is the same as input in Dotnet doesn't make much sense if the input `DateTime` is local.

Zaid-Ajaj:
I'll merge the test now, I will have to write more intergrations tests regarding [#1374](https://github.com/fable-compiler/Fable/issues/1374)

PR #31, opened 2018-03-26
Do not send too many null arguments to server

Zaid-Ajaj:


Fable.Remoting.Client/Proxy.fs
```diff
@@ -197,7 +197,7 @@ module Proxy =
                     let returnType = asyncOfreturnType.GenericTypeArguments.[0]
                     let fieldName = fst field
                     let normalize n =
-                        let fn = proxyFetch state typeName fieldName returnType
+                        let fn = proxyFetch state typeName fieldName returnType (funcTypes.Length - 1)
```
Nhowka: `(funcTypes.Length - 1)` is the same as the argument `n`

Zaid-Ajaj: Fixed! thanks :heart:

PR #33, opened 2018-03-28
End to end testing: turning existing integration tests into unit tests

Zaid-Ajaj:
Testing the whole stack with selenium. Just run:
```
./build.sh IntegrationTests
```
Will be added to the CI pipeline soon.

Needs more polishing when it comes to the results, you will only see something like this:
```
Passed: 54
Failed: 0
```


Nhowka:
Looks like it failed in something other than the tests... The output seems ok
```
Passed: 54
Failed: 0
```

Zaid-Ajaj:
Yeah it seems that selenium sends a request to the firefox driver everytime you ask it for the text content, it threw an exception because I asked for the value after closing the driver, I think it is fixed now

Nhowka:
Now it should be easier to detect those pesky bugs that keeping creeping into the releases! 😄

Zaid-Ajaj:
Awwh yeahh, it is working!!! 



No more bugs hopefully :heart:

Issue #34, opened 2018-03-29
Make HttpContext available from ErrorHandler


Zaid-Ajaj:
Instead of 
```
Exception -> RouteInfo -> ErrorResult
```
I think it would be much nicer to have 
```
Exception -> HttpContext -> ErrorResult
```
There are many reasons why more context information is needed, for example, the `ErrorHandler` is where most people put their logging logic when things go wrong. When that happens, being able to trace back the request using the context is extremely convenient.

@Nhowka What do you think?

Related:
 - [Suave.SerilogExtensions](https://github.com/Zaid-Ajaj/Suave.SerilogExtensions) easy logging integration with Fable.Remoting.Suave

Nhowka:
I was using the Serilog extension when investigating the async bug. I think that is a great idea, but it would be a breaking change code-wise. Are we ready for it? 😋

Zaid-Ajaj:
We could add the `HttpContext` as a field to `RouteInfo` then we won't be breaking anything, so that's another option

Nhowka:
Perfect! That way it's safe and nobody will be sad! 😙

PR #36, opened 2018-03-29
Add HttpContext field to RouteInfo

Zaid-Ajaj:
Addresses #34


Zaid-Ajaj:
With this addition in place, [Logging](https://zaid-ajaj.github.io/Fable.Remoting/src/logging.html) has been documented, only for Suave though (for now)

PR #38, opened 2018-04-09
Update the docs with the correct keyword for setting the auth token in the remote builder

marcpiechura:


Zaid-Ajaj:
Oops, my bad! Thanks for the correction :wink:

Zaid-Ajaj:
Docs site updated! thanks again :smile:

PR #39, opened 2018-04-12
Add callback to acquire a token per request

marcpiechura:
A example usage, based on [ADAL.js](https://github.com/AzureAD/azure-activedirectory-library-for-js) and [Fable.Import.Adal](https://github.com/colinbull/Fable.Import.Adal) would look like this.

```fsharp
// Adal.fs
let private _authContext = 
    let config = createEmpty<adal.Adal.Config>
    config.clientId <- "xxxxxx"
    config.tenant <- Some "xxxxx"
    config.postLogoutRedirectUri <- Some (Browser.window.location.origin + "/logout")
    config.cacheLocation <- Some "localstorage"
    config.redirectUri <- Some (Browser.window.location.origin)
    adal.Adal.createAuthContext(config)

let init () =    
    _authContext.handleWindowCallback()
    let user = _authContext.getCachedUser()
    if isNull user then _authContext.login() 
    else ()
    
let acquireToken() =
  let acquireTokenAsync success error = 
    _authContext.acquireToken(_authContext.config.clientId, (fun errorMsg token -> 
      if isNull errorMsg || errorMsg = "" 
      then success token
      else error (exn errorMsg)    
      None))

  let token = _authContext.getCachedToken _authContext.config.clientId  
  if isNull token || token = "" then Promise.create acquireTokenAsync else Promise.lift token 
  |> Promise.map(fun token -> "Bearer " + token)

// Client.fs

Adal.init()

let server = Proxy.remoting<IServer> {         
      with_token_callback Adal.acquireToken
      with_builder (sprintf "/api/%s/%s")
   }   
```

### DIFF
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -6,13 +6,19 @@ open Fable.Core
 open Fable.Core.JsInterop
 open Fable.PowerPack.Fetch
 
-module Proxy =  
+module Proxy =
     type ErrorInfo = {
         path: string;
         methodName: string;
         error: string;
         response: Response
     }
+
+    type TokenCallbackOption =
+        | Lifetime
+        | OnError of retries:int
+        | EveryRequest
+
     /// On Unauthorized error handler (for backward compatibility)
 
     let mutable private authHandler : Option<string option -> unit> = None
@@ -34,8 +40,6 @@ module Proxy =
     /// When a forbidden error is thrown on the server, this handler intercepts that error along with the optional authorization string information
     let onForbiddenError (handler: string option -> unit) =
         forbiddenHandler <- Some handler
-    /// Cached empty token as default value if neither "with_token" or "with_token_callback" are set
-    let private emptyToken = Promise.lift None
 
     type ResponseContext = {
         Body          : string
@@ -53,7 +57,8 @@ module Proxy =
            CustomHeaders          : Map<string, HttpRequestHeaders list>
            Endpoint               : string option
            Headers                : HttpRequestHeaders list
-           TokenCallback          : (unit -> Fable.Import.JS.Promise<string option>)
+           Authorization          : string option ref
+           TokenCallback          : (TokenCallbackOption * (unit -> Fable.Import.JS.Promise<string option>)) option
            Builder                : (string -> string -> string)
         }
         with
@@ -68,7 +73,8 @@ module Proxy =
                                             ContentType "application/json; charset=utf8"
                                             Cookie Fable.Import.Browser.document.cookie
                                             ]
-                   TokenCallback         = (fun _ -> emptyToken)
+                   Authorization         = ref None
+                   TokenCallback         = None
                    Builder               = sprintf ("/%s/%s")
                 }
     [<Emit("$2[$0] = $1")>]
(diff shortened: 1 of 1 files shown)


Fable.Remoting.Client/Proxy.fs
```diff
@@ -51,6 +53,7 @@ module Proxy =
            CustomHeaders          : Map<string, HttpRequestHeaders list>
            Endpoint               : string option
            Headers                : HttpRequestHeaders list
+           TokenCallback          : (unit -> Fable.Import.JS.Promise<string>)
```
marcpiechura: Since a cached token is only a `Promise.lift` away I figured I could simplify the API even further.

Fable.Remoting.Client/Proxy.fs
```diff
@@ -65,6 +68,7 @@ module Proxy =
                                             ContentType "application/json; charset=utf8"
                                             Cookie Fable.Import.Browser.document.cookie
                                             ]
+                   TokenCallback         = (fun _ -> emptyToken)
```
marcpiechura: Since we need to use some default `Promise` for the expression later on I thought I can simply use a default callback instead of `option`

Fable.Remoting.Client/Proxy.fs
```diff
@@ -148,12 +152,15 @@ module Proxy =
                [ box arg0;box arg1;box arg2;box arg3;box arg4;box arg5;box arg6;box arg7;box arg8;box arg9;box arg10;box arg11;box arg12;box arg13;box arg14;box arg15 ] 
                |> List.take typeCount
             promise {
+                let! token = options.TokenCallback()
+                let authHeader = if isNull token || token = "" then [] else [Authorization token]
```
marcpiechura: Not sure if we should use `string option`  as result of the `Promise`  instead

Fable.Remoting.Client/Proxy.fs
```diff
@@ -257,7 +264,11 @@ module Proxy =
             /// Sets an authorization string to send with the request onto the Authorization header.
             [<CustomOperation("with_token")>]
             member __.WithToken(state,token) =
-                {state with Headers = (Authorization token)::state.Headers}
+                {state with TokenCallback = (fun _ -> Promise.lift token)}
+            /// Sets a callback which is invoked on every request to acquire a authorization string which will be set onto the Authorization header.
+            [<CustomOperation("with_token_callback")>]
+            member __.WithTokenCallback(state,callback) =
+                {state with TokenCallback = callback}
```
marcpiechura: Do I also need to add something to the `createSecureWithEndpointAndBuilderImpl` ?

Fable.Remoting.Client/Proxy.fs
```diff
@@ -148,12 +152,15 @@ module Proxy =
                [ box arg0;box arg1;box arg2;box arg3;box arg4;box arg5;box arg6;box arg7;box arg8;box arg9;box arg10;box arg11;box arg12;box arg13;box arg14;box arg15 ] 
                |> List.take typeCount
             promise {
+                let! token = options.TokenCallback()
```
marcpiechura: Not sure how and which errors I should handle here

Fable.Remoting.Client/Proxy.fs
Nhowka: Isn't it better to have a `string option` as return instead of a `string`? So we wouldn't need to treat the case of empty string and be more consistent.

Fable.Remoting.Client/Proxy.fs
```diff
@@ -148,12 +152,15 @@ module Proxy =
...
+                let authHeader = if isNull token || token = "" then [] else [Authorization token]
+
                 // Send RPC POST request to the server
                 let requestProps = [
                     Body (unbox (toJson data))
                     Method HttpMethod.POST
                     Credentials RequestCredentials.Sameorigin
-                    requestHeaders (customHeaders@options.Headers)
+                    requestHeaders (customHeaders@options.Headers@authHeader)
```
Nhowka: Wouldn't that possibly override a previously defined auth header?

Fable.Remoting.Client/Proxy.fs
marcpiechura: You're reviewing faster than I can add my comments :) I asked myself the same,  see one of the comments below, so I'll change it to string option then.

Fable.Remoting.Client/Proxy.fs
marcpiechura: Wouldn't that end in two auth headers since `options.Headers` is a `list` and not a `set` ? Not sure if that's better though ;) 
It wasn't handled before either but I could check if an auth header is already provided and if so ignore the token.

Nhowka: Probably replacing it would be better, but I think we should make the authorization a part of the option again. Should be easier to handle.

marcpiechura: So adding it to the `Headers` when `with_token` is called as before  + adding or replacing an existing token from the `Headers`  when `callback` returns `Some` ? Maybe we should make the callback an option then too? In order to not await a promise when `with_token_callback` isn't called.

Nhowka: I did some changes for it to be an option too, also some changes to get the token once, every time (the default) or only on error.

marcpiechura:
@Nhowka  Works like a charm 👍

Zaid-Ajaj:
Awesome work @marcpiechura @Nhowka this is really interesting, thanks a lot :heart: I will merge now and see how I can write some tests for this PR.

Nhowka:
I would like the retrying feature to be used in normal context too, but I don't know if we should remove the `retries` from `OnError` risking it getting a default of 0 and don't work as it should. Maybe using one on the options and this one when on error will be the best course.
 
Agree with the naming and API usage.

PR #40, opened 2018-04-13
Configurations for retrying and name changes

Nhowka:
I chose to keep `OnError n` because could have cases where the token has a higher lifetime but be scarce on how many they can be generated. This way the user could only get a new one when necessary.


Zaid-Ajaj:
This looks great, will publish soon 😘

PR #42, opened 2018-04-21
Fix regression introduced with case label rename.

janno-p:


Nhowka:
Wonder how CI was green before that... Thank you for fixing it!

Zaid-Ajaj:
> Wonder how CI was green before that

Because I should have written some tests/docs for them but didn't have the time to :(

Nhowka:
Don't worry, these new features were the most niche we ever had, but I feel like stuff like that should don't even compile. You are doing a great job on all your projects! 😉

Zaid-Ajaj:
Thank you @Nhowka  :kissing_heart: appreciate it!  

>  I feel like stuff like that should don't even compile

I think it is because the unknown case was considered a variable name that it compiled correctly

@janno-p Client package 2.5.1 is published, should be available soon from Nuget

janno-p:
Exactly, it actually produced some annoying warnings in webpack console: first about unconventional label (with uppercase first letter), and since it matched every case, there were also two warnings about patterns wich will be never matched.

Otherwise, great, will update soon.

PR #46, opened 2018-05-09
Allow to pass custom HttpClient to DotnetClient proxy.

irium:
This allows using Fable.Remoting with ASPNET.Core's TestServer and TestClient:
```
    let server = new TestServer(WebHostBuilder().UseStartup<Startup>()
    let client = server.CreateClient()
    let builder = ...
    let proxy = Proxy.custom<'T> builder client
```

### DIFF
--- a/Fable.Remoting.DotnetClient/Proxy.fs
+++ b/Fable.Remoting.DotnetClient/Proxy.fs
@@ -6,73 +6,74 @@ open System.Net.Http
 
 [<RequireQualifiedAccess>]
 module Proxy =
-    
+
     open Patterns
     let private converter = FableJsonConverter()
 
     /// Parses a JSON iput string to a .NET type using Fable JSON converter
-    let parseAs<'t> (json: string) = 
+    let parseAs<'t> (json: string) =
         JsonConvert.DeserializeObject<'t>(json, converter)
-    
+
     /// Sends a POST request to the specified url with the arguments of serialized to an input list
-    let proxyPost<'t> (functionArguments: obj list) url client auth = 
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Nice addition, thanks a lot! already pushed and published in the next version 1.2.0 :smile:

irium:
Thank you too! Because I really need this and currently using home-grown version of it. But it's always better to have it "out of the box" :)

Zaid-Ajaj:
It is refreshing to see someone *actually* using this thing :smile:

Zaid-Ajaj:
File upload/download could be added to the library but I would rather limit the scope of problems the library tries to solve. If you can handle file/upload with a couple lines of code and you happen to need that only once or twice, then a library feature is a bit overkill for that.

irium:
Agree. But some sort of recipe in the docs wouldn't hurt? ;)

I'm still curious about how it could be done : pass it as byte array? But because of JSON-encoding built-in it is wasteful in terms of traffic :(

Zaid-Ajaj:
@irium Indeed passing it as a byte array is a big waste! it doesn't make sense from the proxy's point of view because we are only using `application/json` as the MIME-type for data exchange, so to make it work we would use a special request with `multipart/form-data` for handling files. But again, doing this with "raw" http without abstractions is the way to go IMO. 

BTW added the docs to your PR [here](https://zaid-ajaj.github.io/Fable.Remoting/src/dotnet-http-client.html), let me know if you want add/change anything :smile:

irium:
Docs seems good to me.

Issue #47, opened 2018-05-29
Stack overflow when deserialising int64


csmager:
This is a remarkably simple repro, but it took a *long* time to track down. I'd set up some endpoints and one kept killing the server process with `StackOverflowException` with no hint of what or where it was failing.

It turns out (possibly a separate issue) that the client will serialise a long / int64 like this:
```js
{low: 5, high: 0, unsigned: false}
```
And deserialising this using `Fable.Remoting.Json.FableJsonConverter` will result in a `StackOverflowException`.

I've put together a [small repro](https://github.com/csmager/fable-remoting-repro) contrasting the behaviour of the converter here and the one in [`Fable.JsonConverter`](https://github.com/fable-compiler/Fable/tree/master/src/dotnet/Fable.JsonConverter). The last 3 all cause the same issue.
```fsharp
printfn "Fable.JsonConverter"
parse<int> fable "\"5\""
parse<int> fable "\"+5\""
parse<int> fable "5"
parse<int> fable "\"text\""
parse<int64> fable "\"5\""
parse<int64> fable "\"+5\""
parse<int64> fable "5"
parse<int64> fable "\"text\""
parse<int64> fable "{low: 41, high: 0, unsigned: false}"

printfn "--"

printfn "Fable.Remoting"
parse<int> remoting "\"5\""
parse<int> remoting "\"+5\""
parse<int> remoting "5"
parse<int> remoting "\"text\""
parse<int64> remoting "\"5\""
parse<int64> remoting "\"+5\""
parse<int64> remoting "5"
parse<int64> remoting "\text\""
parse<int64> remoting "{low: 41, high: 0, unsigned: false}"
```
The output is
```
Fable.JsonConverter
Success: "5" -> 5
Success: "+5" -> 5
Success: 5 -> 5
Failed: "text", Could not convert string to integer: text. Path '', line 1, position 6.
Success: "5" -> 5L
Success: "+5" -> 5L
Success: 5 -> 5L
Failed: "text", Input string was not in a correct format.
Failed: {low: 41, high: 0, unsigned: false}, Expecting int64 but got StartObject
--
Fable.Remoting
Success: "5" -> 5
Success: "+5" -> 5
Success: 5 -> 5
Failed: "text", Could not convert string to integer: text. Path '', line 1, position 6.
Success: "5" -> 5L
Success: "+5" -> 5L
Process is terminating due to StackOverflowException.
```
[...]

PR #48, opened 2018-05-29
Long serialization

Zaid-Ajaj:
Addresses #47


Zaid-Ajaj:
@Nhowka @csmager Do you guys know a simple way of converting this:
```
type InternalLong = { high : int; low: int; unsigned: bool }
```
To a simple `int64` value? It seems that this is the only blocker for the issue and honestly I have no idea how this exactly works

Zaid-Ajaj:
[this block of code](https://github.com/Zaid-Ajaj/Fable.Remoting/blob/2cae43125e4b721b26f1ff69dce9a46bee7b6d2c/Fable.Remoting.Json/FableConverter.fs#L226-L229) needs reviewing and it is where the issue can be solved

Zaid-Ajaj:
Ok I think this solves the issue:
```fs
let lowBytes = BitConverter.GetBytes(internalLong.low)
let highBytes = BitConverter.GetBytes(internalLong.high)
let combinedBytes = Array.concat [ lowBytes; highBytes ]
let combineBytesIntoInt64 = BitConverter.ToInt64(combinedBytes, 0)
```
Adding more tests for large values (max, min)

PR #49, opened 2018-06-06
[WIP] Websocket version

Nhowka:
First draft of an untested websocket version. Most of the features gained via headers is lost, so not sure if it is of any worth.


Zaid-Ajaj:
Hi @Nhowka this looks quite nice, but isn't this something solved by your library [Elmish.Remoting](https://github.com/Nhowka/Elmish.Remoting) and both can be used at the same time in suave/giraffe/saturn application?

Nhowka:
Not out of the box. With Elmish.Remoting everything must travel through messages and it has a implicit state, migrating from Fable.Remoting would be too much effort. This one just changes the transport of the data from a POST request to a WebSocket frame, but everything except the http side feels the same.

Zaid-Ajaj:
> This one just changes the transport of the data from a POST request to a WebSocket frame, but everything except the http side feels the same.

Although it is nice to have, I don't see why someone would want to change their POST requests to WebSocket frames, please correct me if I am wrong on this point, but does it add anything? Also won't the server now be needing to keep track of connected clients? 

> Not out of the box. With Elmish.Remoting everything must travel through messages and it has a implicit state, migrating from Fable.Remoting would be too much effort

I am assuming that if someone is using Elmish.Remoting then they *want* stateful chatroom-like architecture where this library is completely stateless, so again, please correct me if I am wrong but aren't these mutually exclusive requirements? 

To be honest (and I would like to hear your opinion on this as well) I would rather limit the feature set of Fable.Remoting but have the currents features: "forget http + json + routing and think functions + types" be done really well like it is doing now, features I *do* want to add are stuff like supporting Fable 1 and Fable 2 at the same time and different ways of exposing the an existing protocol to third party clients

Nhowka:
> does it add anything?

To be honest, not much. Maybe lower latency as it won't have to connect again for every function call, just connecting once and then being sending the arguments and waiting for the result.

> Also won't the server now be needing to keep track of connected clients?

Not really. As stateless, just having the loop for each user is fine. No client knows of each other and the server just answers through the same mailbox it got the message.

> aren't these mutually exclusive requirements?

Yes. This just enables the RPC features from Fable.Remoting through a different channel so the user can use a websocket connection for stateless things.

> features I do want to add are stuff like exposing an existing protocol to third party clients

This makes it even harder. Probably could be used when people don't want to expose a rest API and want to have a single point of exposure.

Being really honest, this would just offer a simpler interface and the user will be able to see all the data sent and received on just one place. Also if the function takes too long to answer, it won't keep an extra connection open, it will just receive the data through the existing connection when the result is ready.

I can't think on anything besides that. I can't think about much of the cons either. The main con I can think of is that it's redundant and will have an extra argument to indicate where the client should transfer the answer to, attenuated by the belief that it will be shorter than the headers. Also wouldn't work with cached answers with service workers and don't know if it could be compressed.

Zaid-Ajaj:
> To be honest, not much. Maybe lower latency as it won't have to connect again for every function call, just connecting once and then being sending the arguments and waiting for the result.

This makes me even more reluctant to adding it, although I am not an expert on latency optimizations, connecting to the server isn't that significant if you take into account the latency caused by CPU-bound work and network IO right?

 > This makes it even harder. Probably could be used when people don't want to expose a rest API and want to have a single point of exposure.

Making the requests harder to replicate manually doesn't solve the problem of exposing the api from a single point, this is more an issue of building secure api's don't you think? 

My point of different ways to expose the api is like building a UI from the type definitions (ideally as an open api spec) to easily test server calls but this is for the future 

Also we would need to fully (integration) test this, maintain it and also making the dotnet client compatible, personally I think it is not worth it if there ins't that much added value and would rather not add it here.

Nhowka:
It's ok, it's mainly a port of what I did before, so not really a wasted effort. Is it ok to leave the branch in case someone wants to test it?

PR #51, opened 2018-06-27
Using parameter instead of overriding

Nhowka:
If weren't by the custom handlers it could be generalized even more. Is that part any useful?


Zaid-Ajaj:
> Is that part any useful?

the custom handlers part you mean or the parameter vs overriding? The former I am not sure, I have mixed feelings on the usefulness of these and on the latter, static functions seem nicer, even though it is just a cosmetic change

Zaid-Ajaj:
I vote for removing custom handlers along with the contextual handlers, it seemed like a good idea at first, but I think generalizing this library to support more http *with* rpc is not what we should be spending our, what do you say (should I open another issue for this)?

Nhowka:
Closing in favor of #53

PR #53, opened 2018-06-28
Lessen code duplication

Nhowka:
Just to show a possible modification that can be done if we drop the custom handlers that are adding a lot of complexity.

### DIFF
--- a/Fable.Remoting.Middleware/FableMiddlewareAdapter.fs
+++ b/Fable.Remoting.Middleware/FableMiddlewareAdapter.fs
@@ -12,9 +12,10 @@ open FSharp.Control.Tasks
 module FableMiddlewareAdapter =
 
   type BuilderOptions<'ctx> with
-    member this.WithBuilder(b) = { this with Builder = b }
-    member this.WithLogger(l) = { this with Logger = Some l }
-    member this.WithErrorHandler(eh) = {this with ErrorHandler = Some eh}
+    member this.WithBuilder b = { this with Builder = b }
+    member this.WithLogger l = { this with Logger = Some l }
+    member this.WithErrorHandler eh = { this with ErrorHandler = Some eh }
+    member this.AddCustomHandler meth ch = { this with CustomHandlers = this.CustomHandlers |> Map.add meth ch }
 
   type RemotingOptions = BuilderOptions<HttpContext>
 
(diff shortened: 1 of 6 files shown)


Zaid-Ajaj:
This made my day :heart:

Nhowka:
I probably can put the custom handlers back without adding complexity back to the server implementations, but we should?

Zaid-Ajaj:
I believe they should be removed, I know, I have been talking a lot about removing stuff, I just want to tidy up the library and get rid of half solutions to problems we shouldn't be concerned with. If someone really wants to use custom handlers and, then why bother with Fable.Remoting in the first place, it is very counter-productive, especially for users, don't you agree?

Nhowka:
I'll add them again just to be sure that it didn't break any test, but after that, we could make it go away for good.

Nhowka:
The build is green. Honestly, I'm proud of the result but not sure if that would make it much easier to create new server implementations. Suave and Giraffe are too much alike. Maybe is a good time to do the Nancy one just to be sure if it will work.

Zaid-Ajaj:
> Maybe is a good time to do the Nancy one just to be sure if it will work

Actually, it will be a lot simpler than you think, because all we need to do is write an aspnet core middleware and someone will be able to use Remoting anywhere, whether it is Nancy, Giraffe or even Freya etc.
```fs
app.UseNancy()
      .UseGiraffe(giraffeApp)
      .UseFreya(freyaApp)  
      .UseRemoting(musicStore, options) // something like this
```
I still haven't given this much thought, but it shouldn't be a problem, even if we were to implement it from scratch with a little bit of code duplication so don't worry about this being *not very extensible*

Zaid-Ajaj:
See how giraffe implements [middleware](https://github.com/giraffe-fsharp/Giraffe/blob/master/src/Giraffe/Middleware.fs) for reference. 

We could be extremely lazy and use giraffe under the hood 
```fs
type IApplicationBuilder with 
  member this.UseRemoting<T>(T serverImpl) = 
    let webApp = remoting serverImpl {()} 
    this.UseGiraffe webApp 
```
Nah, just kidding :smile:

Nhowka:
@Zaid-Ajaj Did what can become a middleware, but didn't test it yet...

Fable.Remoting.Middleware/FableMiddlewareAdapter.fs
```diff
@@ -0,0 +1,52 @@
...
+    member this.WithLogger(l) = { this with Logger = Some l }
+    member this.WithErrorHandler(eh) = {this with ErrorHandler = Some eh}
+
+  type RemotingOptions = BuilderOptions<HttpContext>
+
+  type FableRemotingMiddleware(next          : RequestDelegate,
+                               implementation: obj,
+                               options       : RemotingOptions) =
+
+      do if isNull next then raise (ArgumentNullException("next"))
```
irium: Just a little improvement: `raise (ArgumentNullException("next"))` could be written as `nullArg "next"`

Nhowka: Done, thanks!

Zaid-Ajaj:
@Nhowka Looks like it fits nicely with what is already done in `Server` :smile: You can merge and I can do the testing, since this is asp.net core, we can easily do the [integration testing using the dotnet client](https://zaid-ajaj.github.io/Fable.Remoting/src/dotnet-http-client.html)

PR #67, opened 2018-08-03
C*#* and F*#* typo?

AkosLukacs:
Github renders as `C*#*` and `F*#*` (with the stars). Not 100% sure about this, because looks good on the docs site...
image


Zaid-Ajaj:
It is not a type, the docs aren't meant to be viewed from raw source (on github) but rather from the generated docs site. Using `C#` on the title doesn't work because markdown thinks it is the start of the heading so I add some formatting with `*#*

AkosLukacs:
Oops, sorry

PR #68, opened 2018-08-10
Automagic docs generation

Zaid-Ajaj:
This PR adds the ability to generate documentation UI on the fly from protocol definition, similar to swagger.
here is an example:
```fs
// Shared protocol definition
type IServerApi = { getLength : string -> Async<int> }

let server = { getLength = fun input -> async { return input.Length } } 

let docs = Docs.createFor<IServerApi>() 

let serverApiDocs = 
    Remoting.documentation "Server docs" [
        docs.route <@ fun api -> api.getLength @>
        |> docs.description "Returns the number of characters the input string has"
        |> docs.example <@ fun api -> api.getLength "one example" @> 
        |> docs.example <@ fun api -> api.getLength "or many examples" @>   
    ]

let serverApi = 
  Remoting.createApi()
  |> Remoting.fromValue server 
  |> Remoting.withDocs "/api/docs" serverApiDocs 
  (* ... *) 
```
The option `Remoting.withDocs` does a couple of things:
- When users navigate to `"/api/docs"`, they will see the documentation UI generated from an embedded docs app
- Another internal route is added `OPTIONS /api/docs/$schema` which returns a JSON representation of the documentation, the embedded docs app retrieves this schema making it available on the client 

docs

## This PR is still WIP 

Implemented
 - [x] Docs API 
 - [x] Schema generation
 - [x] Exposing schema through HTTP
 - [x] Docs App
     - [x] Shows available remote functions and their description, route etc. 
     - [x] Can navigate between functions  
     - [x] Can run `GET` requests 
     - [x]  Can run `POST` requests 
     - [x]  Can write request body from text area as input for `POST` requests
     - [x]  Make the route function details "sticky" (nice to have)
     - [x]  Use added examples from the UI
 - [x] Code clean up 
 - [ ] ~~UI tests for the docs~~


StachuDotNet:
(this is *very* cool!)

Zaid-Ajaj:
Thanks @StachuDotNet I am glad you like it :smile: here is a gif of the final version 

docs2 0

StachuDotNet:
Thank *you*. I could see myself using this in a production capacity in the coming weeks.

brettrowberry:
Is the JSON representation OpenAPI compliant?

reinux:
> That said, I could still add another route that generates an OpenAPI schema, but I wonder if it provides more value than what the docs app already gives you.

I'd gladly take the compromises, as it would mean that I could prototype the frontend in F#, and then bring people in to work with whatever frontend stack they're already comfortable working in.

PR #72, opened 2018-08-19
Remove unneeded log

nojaf:


Zaid-Ajaj:
Oeps my bad 😅 published in beta-003

PR #74, opened 2018-09-04
Error when passing Map with DU as Keys in Fable 2 client.

irium:
I've added two test cases:
- `IServer.echoConfigMap`  - "canonical" form. Error thrown is somewhat different than my real case, which I've implemented as 2nd test case:
- `IServer.returnConfigMap`. Error at the client side is because it compares key name in quotes ("\"K1\"" in the test) with key name without them ("K1") and didn't find corresponding key.


irium:
Closing this in favor of https://github.com/Zaid-Ajaj/Fable.SimpleJson/pull/3

PR #79, opened 2018-10-06
Fixed combineRouteWithBaseUrl

mvsmal:
Since `route` always starts with `/` according to `RouteBuilder = sprintf ("/%s/%s")`, `combineRouteWithBaseUrl` always resulted with a final url like `http://baseurl//IMusicStore/allAlbums`.
This fix avoids duplicated slash.


Zaid-Ajaj:
@mvsmal Thanks a lot for spotting this!

PR #86, opened 2018-10-29
Remove powerpack in favor of simple XMLHttpRequest

### LINKED ISSUE #80: Q: Ending up with two libraries to serialize things.
Hi, I'm using `Fable.Remoting.Client` and I noticed it has a dependency on `Fable.PowerPack` and `Fable.SimpleJson`. If you look closer `Fable.PowerPack` has a dependency on `Thoth.Json`.

So I guess in my bundle I end up with two libraries that serialize json. I'm having some mixed feelings about this. I'd like to hear your thoughts on this.

//cc @MangelMaxime

Zaid-Ajaj:
Fixes #80 and simplifies the overall proxy implementation 

### Breaking changes:

`CustomHeaders` in type `RemoteBuilderOptions` are now of type `(string * string) list` 

`Response` in `ProxyRequestException` is simplified to:
```fs
type HttpResponse = {
    StatusCode: int 
    ResponseBody: string
}
```
instead of using the full `Response` type from PowerPack

### Notes 

@Elonon I am not sure what the equivalent is to `Crendentials: Same-origin` when I am using `XMLHttpRequest`, any ideas?


Elonon:
``XMLHttpRequest`` will send ``httpOnly`` cookies by default as long as it is the same origin. For other domains, you must set ``withCredentials`` = ``true``.

See [here ](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials)for more info.

Elonon:
``httpOnly`` cookies still aren't available in ``document.cookie``, but you are correct that to match your previous set up, you don't need to set ``withCredentials`` to ``true``. I mentioned it because maybe you want to provide an option if a user needs it. You should not set it to ``true`` by default at any rate.

Zaid-Ajaj:
@Elonon That would be awesome! Thanks a lot :smile: send a PR anytime you like and I will review

PR #112, opened 2019-02-23
Documentation updates - removed reference to deprecated Saturn function and tweaked suggested webpack "devServerProxy" config

drk-mtr:
- Updated `saturn.md`, `README.md` and `dependency-injection.md` to reference `use_router` rather than the deprecated `router`.
- Updated `client.md` to use what I believe is the correct syntax to help avoid http-proxy-middleware (https://github.com/chimurai/http-proxy-middleware) error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -209,7 +209,7 @@ let webApp : HttpHandler =
 
 let app = application {
     url "http://127.0.0.1:8083/"
-    router webApp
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj:
Thanks a lot @drk-mtr! I just published a new version of the docs

PR #115, opened 2019-03-25
Cancel Async if the Request is aborted

0x53A:
I have a (relatively) expensive operation in one of my server functions.

If the client cancels the request (because it timed out client-side, or because the user closed the browser) I would like to be notified.

There is a CancellationToken exposed in the HttpContext that fires if the client aborts: https://github.com/aspnet/HttpAbstractions/issues/197#issuecomment-279129671

This should just forward that token to Async.

### DIFF
--- a/Fable.Remoting.AspNetCore/Middleware.fs
+++ b/Fable.Remoting.AspNetCore/Middleware.fs
@@ -87,7 +87,7 @@ module internal Middleware =
       fun (next : HttpFunc) (ctx : HttpContext) -> 
         task {
             Diagnostics.runPhase logger func.FunctionName
-            let! functionResult = Async.StartAsTask (Async.Catch (DynamicRecord.invokeAsync func impl args)) 
+            let! functionResult = Async.StartAsTask( (Async.Catch (DynamicRecord.invokeAsync func impl args)), cancellationToken=ctx.RequestAborted)
             match functionResult with
             | Choice.Choice1Of2 output -> 
                 ctx.Response.StatusCode <- 200
--- a/Fable.Remoting.Giraffe/FableGiraffeAdapter.fs
+++ b/Fable.Remoting.Giraffe/FableGiraffeAdapter.fs
@@ -53,7 +53,7 @@ module GiraffeUtil =
       fun (next : HttpFunc) (ctx : HttpContext) -> 
         task {
             Diagnostics.runPhase logger func.FunctionName
-            let! functionResult = Async.StartAsTask (Async.Catch (DynamicRecord.invokeAsync func impl args)) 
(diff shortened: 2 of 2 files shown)


0x53A:
My scenario is already possible using ``Remoting.fromContext`` instead of ``Remoting.fromValue``, do you think this would make sense anyway?

If yes I'd add something similar to the other entry points (suave, etc) if they also expose a cancellation handler,

0x53A:
Hm I don't quite understand the travis error:

grafik

Zaid-Ajaj:
Looks like a good default choice for people using `Remoting.fromValue`, I like it

> If yes I'd add something similar to the other entry points (suave, etc) if they also expose a cancellation handler,

Yes please

> Hm I don't quite understand the travis error:

I restarted the build, but anyways seems safe to ignore since all integration tests passed :smile:

Zaid-Ajaj:
Nope, looks like Travis was right:

Untitled

0x53A:
yeah, that happened because I was too lazy to clone and edited it _blind_ in the github UI.

So apparently AppVeyor doesn't compile the giraffe stuff.

Will add the other two in the evening

Zaid-Ajaj:
> So apparently AppVeyor doesn't compile the giraffe stuff.

Indeed, AppVeyor only runs the end-to-end tests using a Suave server, I probably should change that

> Will add the other two in the evening

Just let me know when this is ready, so I can roll out new packages :smile:

0x53A:
I couldn't find any ConnectionAborted event for suave, but I'm not familiar with it. /cc @haf

I did add it to the raw AspNetCore server.

Also, I'm a little disappointed, the event fires a bit late, maybe 10 seconds or more after I close the browser window. Still better than nothing, but shouldn't it be able to detect a closed tcp connection "instantly"?

Zaid-Ajaj:
> I couldn't find any ConnectionAborted event for suave, but I'm not familiar with it. /cc @haf

I will take a look

> I did add it to the raw AspNetCore server.

Awesome, I will merge and publish Giraffe & AspNetCore adapters first

> Also, I'm a little disappointed, the event fires a bit late, maybe 10 seconds or more after I close the browser window. Still better than nothing, but shouldn't it be able to detect a closed tcp connection "instantly"?

Honestly, I have no idea, maybe this is something to file at the Ap.net core repo

0x53A:
Thanks!

> Honestly, I have no idea, maybe this is something to file at the Ap.net core repo

Yes, it was more of a statement than a question. It's still good to have - I had an issue where the requests all went into a queue, the clients timed out but and the queue continued to grow, forcing a hard reboot of the whole server because the CPU was maxed out and we couldn't even connect remotely to restart the process.

Issue #116, opened 2019-03-26
[Idea] add a high performance path with binary transfer and direct access to an ArrayBuffer


0x53A:
_This is a bit different from the current goal of Fable.Remoting, and may be out-of-scope for you. But I like the programming model of Fable.Remoting. So I wanted to at least ask you for your opinion._

Even though the below suggestion is already detailed, I'm not too married to it. I would just really like _any_ way to transfer big(ish) blobs without the (de)serialization overhead.

----------------------

I have a scenario where I need to transfer big objects.

In the first prototype I serialized them to base64 strings then transfered them over Fable.Remoting. In that case, the underlying web request took ~2 seconds, but the F.R Api call took 20 seconds. So 18 seconds were spent just parsing and deserializing big base64 encoded strings.

Currently I have a secondary Giraffe Webhandler which returns the binary data, and on the client side I do a ``fetch`` and access the ArrayBuffer directly.

With this and other improvements I was able to reduce total time for request+processing from over a minute to 10 seconds.

While not a big issue currently, this has a few downsides:

1) I have two seperate APIs for no reason
2) Duplicated authentication
3) no easy input-parameter handling (currently I have just an int-array in the body, so not much of an issue)

---------

I would like to unify this again with the rest of my API and would propose to add a binary fast-path to Fable.Remoting:

```f#
// Add this to Fable.Remoting
type BinaryBuffer() =

    // Fable Client Side
    member this.GetArrayBuffer() : Async<Fable.Import.JS.ArrayBuffer> = // ...

    // .NET ServerSide
    static member FromStream(s:System.IO.Stream) : BinaryBuffer = // ...
    static member FromArray(a:byte array) : BinaryBuffer = // ...
```

My API record would then look like this:

```f#
type IRemotingApi = {
    // typical request, serialized as json
    GetSomeData: Async<MyData>
    // fast binary response, no serialization & deserialization involved
    Load3DModelsForIDs: int array -> Async<BinaryBuffer>
}
```

On the server, I would create a buffer from either an array or a stream, and on the client I would use it like this:

```f#
let ids = [| 1;2;3 |]
let! response = api.Load3DModelsForIDs(ids)
let! buffer = response.GetArrayBuffer()

// now I can work with the array buffer

// maybe the buffer represents a list of vertices (floats) for my 3d model
let typedArray = JS.Float32Array.Create(buffer)
```

__Downsides:__

* I lose the strong typing, the blob can represent anything
[...]

PR #117, opened 2019-03-26
Implement binary responses that bypass deserialization

Zaid-Ajaj:
Addresses #116 

Implements `byte[]` responses that are read directly as `UInt8Array` and thus bypassing the usual pipeline of converting first to base64 string just to read it back to an array of bytes.

The condition for binary responses is that the output type for a route must be `Async<byte[]>` and that the request includes a dummy header "x-remoting-proxy" so that old proxies and dotnet client stay compatible. 

Added tests for binary routes to new proxy and old dotnet client to make sure these still work.

@0x53A 
The input parameters are still serialized to JSON even they are `byte array`. For only I want to verify the performance gains from bypassing `byte array` deserialization for large arrays

To benchmark the differences, use one route that returns `Async<byte[]>` and another that returns `Async<Option<byte[]>>`. The content is the same but the latter will go through deserialization while the former is converted directly


Zaid-Ajaj:
@0x53A if I start rolling out new packages with the bypass, will you be able to test/benchmark them against your use-cases?

0x53A:
Yeah. Not in my main project (Fable.Core 3 again), but should be easy to extract to a small benchmark programm.

Zaid-Ajaj:
It's settled then, packages are published and will be available when nuget does it's magic. I look forward to hearing your findings :smile: if the performance gains are good and everyone is happy, I can then take a look at optimizing the binary inputs (i.e `byte[]`) to bypass input deserialization on the server (and serialization from the client)

0x53A:
So I've finally had time and created a small benchmark (will upload to github in a few minutes).

-----------------

At first I accidentally ran it with old server and new client.

It seems like this combination does not work correctly - both server functions (slow and fast) return the same byte array, but on the client I receive byte arrays with differing lengths:

grafik


grafik

grafik

So fast path with old server will return a blob that is too big. Reason is probably that it just returns the JSon un-deserialized.

Not sure if that is a scenario you want to support, if so, you'd need to add a header to the response, just like you did with the request.

--------------------

0x53A:
Did you also upload the server nuget? I can't find any pre-release

grafik

0x53A:
Hm, according to the github PR view, 3.5.0 _should_ contain the changes, but my above test (where I thought I had an old server) _was_ with 3.5.0.

0x53A:
repo is here: https://github.com/0x53A/fable-remoting-117

I'm tired and probably just made a silly mistake. Will look again at all this in the morning.

Zaid-Ajaj:
Thanks a lot for the repro, I will check it right away.

> So fast path with old server will return a blob that is too big. Reason is probably that it just returns the JSon un-deserialized.
> Not sure if that is a scenario you want to support, if so, you'd need to add a header to the response, just like you did with the request.

Old server, new client is not a scenario I accounted for. Only the other way around, New server is still compatible with old clients (both Fable and Dotnet). I will fix that in the Fable client

> Hm, according to the github PR view, 3.5.0 should contain the changes, but my above test (where I thought I had an old server) was with 3.5.0.

I didn't push a pre-release because no breaking change was introduced, so 3.5.0 is the most recent version

0x53A:
Giraffe checks x-binary-content
grafik

Client sends x-remoting-proxy

grafik

0x53A:
You are fast, now NuGet is the bottleneck ;D

Zaid-Ajaj:
My bad :( the fix is already published. Again, my integration tests are running with the Suave server where I manually checked the responses and it all worked out but it seems one can never have enough tests

0x53A:
Okay real test:

fast path is 25% smaller and also 20%-30% faster (3 seconds instead of 4)

grafik


grafik

0x53A:
Chrome is a bit faster, but here the fast path is also notacibly faster:

grafik

grafik

As you can see the slow path has a big CPU usage block after the network request, where the fast path doesn't

0x53A:
This looks really good, thank you! 🎆

0x53A:
Just checked, it's not as much, but skipping serialization is also slightly (~100ms) faster on the server:

grafik

Zaid-Ajaj:
I just reproduced these results and . . . wow! Not only the CPU time was reduced, the network latency too, man I am happy now :smile: I guess I will start working on optimizing binary inputs too very soon

On a side note, I am very intrigued to see someone pushing the library to it's limits. If I may ask, what kind of application are you building, is it a game, maybe even a game engine?

0x53A:
Nope not a game. Not public yet, but I'd be happy to send you a DM on twitter or somewhere else.

Zaid-Ajaj:
Sounds like an a cool project! Also DMs are always open ;)

0x53A:
It is cool, Fable is really fun! (They should have kept the name FunScript ;) ) - I did send you one with a link, and we'll present it publically in ~2 weeks. Good night :D 💤

Issue #127, opened 2019-07-30
json serialization is relatively expensive


0x53A:
Hi @Zaid-Ajaj , it's me again :D

I have a remoting api that returns an object graph with a lot of little objects.

![image](https://user-images.githubusercontent.com/4236651/62150770-19288000-b2ff-11e9-8bad-fd0214acb376.png)

Of the total time spent, 5% is my (non-trivial) processing, 95% is json serialization. The total duration is about 10 seconds.

Looking at the lower table shows the possible big wins:

I assume the 11% of ``UnionCaseInfo.get_Name`` could be halved to 5% by just caching ``uci,Name`` into a local.

![image](https://user-images.githubusercontent.com/4236651/62151075-d1eebf00-b2ff-11e9-9716-6e3ca37d72ac.png)

The much bigger win would be speeding up / replacing ``GetUnionFields``.

Maybe be ``UnionCaseInfo`` could be cached, and then the fields could be read by accessing the ``PropertyInfo``s:

```f#
[<Sealed>]
type UnionCaseInfo =
class
// ...
member this.GetFields : unit -> PropertyInfo []
// ...
end
```

Anyway, at the moment this is just idle speculation. I'll see if I can create some benchmark over the next few weeks and a PR, but I really need a vacation tbh.

What do you think? Would such a caching approach make sense?

Zaid-Ajaj:
Hello there, 

Thanks for the detailed report! It would be great to have JSON benchmarks :rocket: preferably in a separate repository and a clone of Fable.Remoting.Json with changes made to it so we can see how they compare to each other instead of starting to mess with the code here with prototypes. 

Do you have a type you can share here with a lot of nested records / fields that generates a 8-10 second delay just for serialization?

At a first naive glance, it looks like the solution could be applying memoization against `GetUnionFields` using equality on the input type, but like you said, this is still speculative. 

> but I really need a vacation tbh.

Yeah me too, man 🏖 

> Would such a caching approach make sense?

I am honestly not sure but it is worth the try, in the past I've tried some caching techniques but didn't yield significant results, I guess we have to test things and look at the benchmark results first

0x53A:
> Do you have a type you can share here with a lot of nested records / fields that generates a 8-10 second delay just for serialization?

https://github.com/0x53A/FableRemotingSerializationTest

It isn't much of a benchmark (yet).

I dumped an object from my dev environment and copied all the type definitions into a new project.

[...]

PR #128, opened 2019-07-31
performance: add a cache for UnionCaseInfo

0x53A:
closes https://github.com/Zaid-Ajaj/Fable.Remoting/issues/127

based on https://github.com/0x53A/FableRemotingSerializationTest/commit/a28e01203a518164367769ef4b4983c741ceb743

image

__Question:__ Is using just a Dictionary safe, or could the instance be shared between Threads?

Edit: Answer: one instance is shared, so need to use ConcurrentDictionary

### DIFF
--- a/Fable.Remoting.Json/FableConverter.fs
+++ b/Fable.Remoting.Json/FableConverter.fs
@@ -144,6 +144,22 @@ type FableJsonConverter() =
         FSharpType.GetUnionCases(t)
         |> Array.find (fun uci -> uci.Name = name)
 
+    let unionCaseInfoCache = Dictionary<Type,string*PropertyInfo array>()
+
+    let getUnionCaseNameAndFields value t =
+        match unionCaseInfoCache.TryGetValue t with
+        | true, (uciName,fieldPropInfos) ->
+            let fields = [|
+                for p in fieldPropInfos -> p.GetValue(value)
+            |]
+            uciName, fields
+        | false, _ ->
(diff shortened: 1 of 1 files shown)


0x53A:
huh, good thing you have tests. It seems the type-based caching doesn't work with struct unions, because all cases share one type.

0x53A:
Same for union cases without fields.

0x53A:
Ok locally tests are green and my use case is still fast:

image
(https://github.com/0x53A/FableRemotingSerializationTest/commit/108a274e1b7b76b1b17a810441d472cb5dc389e2)

Zaid-Ajaj:
Mind blown by these numbers :rocket: really awesome work @0x53A ❤️

Zaid-Ajaj:
Updated and published all (server side) packages, should be available soon on nuget :smile:

0x53A:
Thanks! I updated my app:

image


image

Now it is 44% processing, 54% serialization. Total time is maybe a second or so.

This is really good - for reference, this loads about a days worth of data.

------

And I think this is as good as it can get with Newtonsoft.Json, I couldn't find anything obvious:

image

So if performance were still an issue (it isn't for me), you could investigate [System.Text.Json](https://devblogs.microsoft.com/dotnet/try-the-new-system-text-json-apis/).

Zaid-Ajaj:
Wohhoo awesome results! I doubt we will get a perf issue any time soon if it can handle so much data at once! Although I wonder what one could squeeze out the new JSON APIs

PR #131, opened 2019-08-01
Upgrade infrastructure and restructure unit-tests

Zaid-Ajaj:


Zaid-Ajaj:
The PR deprecates the legacy Remoting.Client v1.x which was used in combination with dotnet-fable

PR #141, opened 2019-11-11
Copy request body to MemoryStream asynchronously

kerams:


Zaid-Ajaj:
Thanks @kerams I have patched and published the Giraffe adapter as of v3.10.0 🚀

PR #146, opened 2019-11-28
Don't print args deserialization exception to stdout

kerams:


Zaid-Ajaj:
Thanks a lot @kerams I have merged and published your patch :rocket: should be available as soon as nuget finishes indexing

PR #147, opened 2019-11-29
Cache implementation record func info

kerams:
This shaves off a millisecond or two on every request ;)


Zaid-Ajaj:
Why not, milliseconds cost 💸 merged and published to nuget!

> Wow, travis even fixed itself 🎉

PR #152, opened 2020-05-12
Fixes #151

### LINKED ISSUE #151: Ability to deserialize private constructors
Here's my failing case:
```
open Newtonsoft.Json

let c = Fable.Remoting.Json.FableJsonConverter()

type String50 = private String50 of string
let s = "[\"String50\", \"onur\"]"

let o = JsonConvert.DeserializeObject<String50>(s,c)
printf "%A" o
```
Fails with 
```
Cannot deserialize the current JSON array (e.g. [1,2,3]) into type 'Program+String50' because the type requires a JSON object (e.g. {"name":"value"}) to deserialize correctly.
```

Removing the `private` fixes the problem. However it should be supported imho.

OnurGumus:


Zaid-Ajaj:
Thanks a lot for filing the issue and for the fix :pray: 

However, I have one concern about the fix. You make union case fields private so that you can constrain how they are created and be able to make assumptions about the data, correct? In case of `String50`, you assume that the data will always have length of 50 or less. 

This fix will *bypass* the creation constrain and malicious users (client apps) will be able to send an invalid JSON such as `[ "String50", null ]` or `[ "String50", <very long string here> ]`. I don't have a problem adding this PR as a temporary fix but then you should be aware of the risks even though the chances are low.

I believe allowing for custom JSON handlers for these types but I am not yet sure how the API will look like, I am open to suggestions of course :smile:

OnurGumus:
@Zaid-Ajaj your point is entirely valid. However it's a bit debatable if we should rely on type safety for security. The primary reasons we use accessors is mostly for bugs. Sure supporting private creates a false sense of security. A developer might falsely assume that the type will never be constructed. However not supporting this feature is like: "Don't put any locks because they are breakable easily". Again it's a matter of opinion, but the convenience of this feature outweighs the potential harm. For proper security you should validate the input no matter what. I have created another pull request, because this alone didn't work for my case.

Zaid-Ajaj:
Alright @OnurGumus, I am sold :smile: let's get this fix in. It will be awesome if you can add a test or two to make sure this behaviour always works as expected and won't break without us noticing.

Zaid-Ajaj:
I added the tests and published all the goodies :100: just update to latest and things should work

PR #163, opened 2020-07-16
Fix incorrect MsgPack serialization caching

kerams:


Zaid-Ajaj:
Merged and published msg pack and everything downstream :rocket: just update anything to latest version and it will pick latest MsgPack implementation

PR #176, opened 2020-08-01
Remove redundant target frameworks

kerams:
Targetting `net462` on top of `net45` is unnecessary because the latter can be used in any application compiled against .NET 4.5 through 4.8.
Targetting `netcoreapp3.1` on top of `netstandard2.0` is unnecessary for the same reason on .NET Core as well as the fact that they don't have a different set of dependencies either (like [this](https://www.nuget.org/packages/Giraffe/4.1.0) where `netcoreapp3.0` drops obsolete ASP.NET packages, symplifying dependency graphs and lock files in .NET Core 3.0+ applications). The only exception here is the MsgPack project, which [conditionally relies](https://github.com/Zaid-Ajaj/Fable.Remoting/blob/3f8e2733bcee99e938bd67300bc50ae9a411d392/Fable.Remoting.MsgPack/Write.fs#L167-L193) on specific .NET Core APIs.

The result of this change are slightly faster builds and slightly smaller Nuget packages.

### DIFF
--- a/Fable.Remoting.DotnetClient/Fable.Remoting.DotnetClient.fsproj
+++ b/Fable.Remoting.DotnetClient/Fable.Remoting.DotnetClient.fsproj
@@ -8,7 +8,7 @@
         <PackageTags>fsharp;fable;remoting;rpc;webserver;json</PackageTags>
         <Authors>Zaid Ajaj</Authors>
         <Version>3.5.0</Version>
-        <TargetFrameworks>net45; net461; netstandard2.0; netcoreapp3.1</TargetFrameworks>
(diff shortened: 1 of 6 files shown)


kerams:
Also maybe gitignore ionide cache? :)

Zaid-Ajaj:
I don't know for sure if we should drop `netcoreapp3.1` because I think (and I am not sure) that it is sometimes needed when using the library in azure functions. I can't find the issue but it was something like that

Zaid-Ajaj:
I guess we will hear about if it really an issue :wink: let's get this in for now

PR #177, opened 2020-08-08
Optimize request proxying

kerams:
- Don't recreate a proxy with TypeShape for every request 🤦‍♂️
- Consider using multiple remoting APIs such as
  ```fsharp
  .UseGiraffe (choose [ deckApiHandler; accountApiHandler; learnApiHandler; otherApiHandler; statsApiHandler ])
  ```
  For `statsApiHandler` requests, implementations for the first 4 handlers would be unnecessarily created if `FromContext` is used, which could be wasteful if those have expensive dependencies. A new `MemoryStream` would be allocated for each of the 4 too, regardless of the fact whether routes match or not.

### DIFF
--- a/Fable.Remoting.AspNetCore/Middleware.fs
+++ b/Fable.Remoting.AspNetCore/Middleware.fs
@@ -88,17 +88,18 @@ module internal Middleware =
                 | Propagate error -> return! setBody (Errors.propagated error) logger next ctx  
         }
 
-    let buildFromImplementation<'impl> (impl: 'impl) (options: RemotingOptions<HttpContext, 'impl>) =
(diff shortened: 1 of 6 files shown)


kerams:
That failure again, hmmm.

Zaid-Ajaj:
Let me give it a try locally

Zaid-Ajaj:
Works locally like a charm 😍 thanks a lot! I published `Server` and its dependants to nuget and it should be available soon. Also published the docs to the Github pages from #178 when you update the docs, you need to run `npm run publish` from the `./documentation` directory :smile:

kerams:
So I should forget the fork and work in a local branch and create PRs from there to master?

Zaid-Ajaj:
Yeah it is easier like that 😄

PR #184, opened 2020-09-24
Add Set<'T> to Fable.Remoting.MsgPack

Shmew:
* Adds `Set<'T>` for (de)serialization in MsgPack
* Added the discussed fix to enable usage of the `Write` module via Fable
* Added some missing test cases for Fable.Remoting (`Set<'T>`)

### DIFF
--- a/ClientV2Tests/webpack.config.js
+++ b/ClientV2Tests/webpack.config.js
@@ -27,7 +27,11 @@ module.exports = function (evn, argv) {
     port: 8081, // where to the run dev-server
     proxy: {
       '/api/*': { // tell webpack-dev-server to re-route all requests from client to the server
-        target: "http://localhost:8080",// assuming the suave server is hosted op port 8080
+        target: "http://localhost:8397",// assuming the suave server is hosted op port 8080
+        changeOrigin: true
+      },
+      '/IAuthServer/*': {
+        target: "http://localhost:8397",// assuming the suave server is hosted op port 8080
         changeOrigin: true
       }
     }
--- a/ClientV2Tests/webpack.config.js
+++ b/ClientV2Tests/webpack.config.js
@@ -27,11 +27,11 @@ module.exports = function (evn, argv) {
     port: 8081, // where to the run dev-server
     proxy: {
       '/api/*': { // tell webpack-dev-server to re-route all requests from client to the server
-        target: "http://localhost:8397",// assuming the suave server is hosted op port 8080
+        target: "http://localhost:8080",// assuming the suave server is hosted op port 8080
         changeOrigin: true
       },
       '/IAuthServer/*': {
-        target: "http://localhost:8397",// assuming the suave server is hosted op port 8080
+        target: "http://localhost:8080",// assuming the suave server is hosted op port 8080
         changeOrigin: true
       }
     }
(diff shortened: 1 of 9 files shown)


Fable.Remoting.MsgPack/Write.fs
```diff
@@ -400,243 +412,269 @@ let serializeObj (x: obj) (out: Stream) =
...
+                    let case, fields = FSharpValue.GetUnionFields (x, t, true)
+                    let fieldTypes = case.GetFields () |> Array.map (fun x -> x.PropertyType)
+                    writeUnion out case.Tag fieldTypes fields) x out
+            elif FSharpType.IsTuple t then
+                let fieldTypes = FSharpType.GetTupleElements t
+                cacheGetOrAdd (t, fun x out -> writeTuple out fieldTypes (FSharpValue.GetTupleFields x)) x out
+            elif t.IsEnum then
+                cacheGetOrAdd (t, fun x -> writeInt64 (box x :?> int64)) x out
+            elif t.IsGenericType then
+                let tDef = t.GetGenericTypeDefinition()
```
kerams, line 619: This is a good refactor. Might be a good idea to pull `t.GetGenericArguments ()` from the 4 branches into outer scope too.

kerams:
The set addition is great, but I'm not sure how I feel about the 2 Fables in the fully qualified of `Write` even though it certainly solves all the problems.

Shmew:
Yeah, I'm not really happy with either of them. So whatever you and @Zaid-Ajaj want works for me.

ClientV2Tests/webpack.config.js
```diff
@@ -27,7 +27,11 @@ module.exports = function (evn, argv) {
     port: 8081, // where to the run dev-server
     proxy: {
       '/api/*': { // tell webpack-dev-server to re-route all requests from client to the server
-        target: "http://localhost:8080",// assuming the suave server is hosted op port 8080
+        target: "http://localhost:8397",// assuming the suave server is hosted op port 8080
```
Zaid-Ajaj, line 30: I don't think we should change this because the port of the server of the integration tests is always random. Instead, when running integration tests server, if not in headless mode, then it should choose 8080 by default:
```fs
if not (Array.contains "--headless" argv) then
  printfn "Starting web server..."
  let config = { suaveConfig with bindings  = [ HttpBinding.createSimple HTTP "127.0.0.1" 8080 ]
  startWebServer config testWebApp
  0
else
  ...
```

Zaid-Ajaj:
The implementation and the tests look great to me. I had one minor comment about configuring the live integration tests, feel free to fix it or just leave it as is (I will fix it later) :smile: the rest looks ready to merge. 

> My assumption is that most people would open Fable.Remoting.MsgPack so it would be Fable.Write.writeObject which didn't seem too bad, but does feel a bit odd.

Yeah that sounds good!

ClientV2Tests/webpack.config.js
Shmew: Yeah I figured that's what it should do, but I couldn't get the live tests to run without changing it to that address ¯\_(ツ)_/¯. I'll change it back though since I'm done testing them anyways. Speaking of which there was one test still failing the live version that didn't fail when running normally which is the auth test. It originally failed because the dev server was missing that path, but after that it was still failing.

kerams: That test never worked for me either so I just ignore it.

Shmew: It works fine (and passes) for the non-live version, if you're referring to the live testing it didn't actually run webpack-dev-server, but I fixed the fsx

PR #186, opened 2020-10-06
Add dotnet client remoting without quotations

Arshia001:
This library comes up in first place in a google search for "F# RPC", and with good reason. However, the dotnet client feels like an after-thought and has always steered me away until now, so I decided to create a similar remote API builder to the one available in the JS client.

I tried to keep the public interface as similar to the JS client as possible, with the exception that a base path is mandatory for the dotnet client. I do believe the implementation is as performant as any normal F# function and should not have any extra overhead once the proxy is built.

I also noticed that setting the authorization header caused the code to choke on the second request onwards, so I fixed that as well.

### DIFF
--- a/Fable.Remoting.DotnetClient/ClientRemoting.fs
+++ b/Fable.Remoting.DotnetClient/ClientRemoting.fs
@@ -127,7 +127,9 @@ module ClientRemoting =
         | Some token -> client.DefaultRequestHeaders.Add("Authorization", token)
         | _ -> ()
 
-        let ctor = t.GetConstructors() |> Seq.exactlyOne
+        match t.GetConstructors() |> Seq.tryExactlyOne with
+        | None -> failwith "Your record type has more than one constructor, probably because you made it [<CLIMutable>]"
+        | Some ctor ->
         let parameters =
             ctor.GetParameters()
             |> Array.map (fun param ->
--- a/Fable.Remoting.DotnetClient/ClientRemoting.fs
+++ b/Fable.Remoting.DotnetClient/ClientRemoting.fs
@@ -89,7 +89,6 @@ module ClientRemoting =
         CustomHeaders: (string * string) list
     }
 
-    // Warning: if routeBuilder returns a path with a leading /, it will override any subpath inside baseUrl
     let createApi (baseUrl: string) =
         {
             RouteBuilder = None
@@ -149,7 +148,7 @@ module ClientRemoting =
                     )
                     |> Seq.toArray
 
-                let route = Uri(options.BaseUri, routeBuilder t.Name param.Name) |> string
+                let route = Uri(options.BaseUri, (routeBuilder t.Name param.Name).TrimStart('/')) |> string
                 if Array.length argTypes = 1 then
                     typedefof<ParameterlessServiceCall<_>>
                         .MakeGenericType(argTypes.[0])
--- a/Fable.Remoting.DotnetClient/ClientRemoting.fs
+++ b/Fable.Remoting.DotnetClient/ClientRemoting.fs
@@ -81,32 +81,34 @@ module ClientRemoting =
             fun f g h -> Proxy.proxyPost<'i> [ box a; box b; box c; box d; box e; box f; box g; box h ] route client isBinarySerialization
 
     type RemoteBuilderOptions = private {
-        routeBuilder: string -> string -> string
-        baseUri: Uri
-        client: HttpClient option
-        authToken: string option
-        isBinarySerialization: bool
-        customHeaders: (string * string) list
+        RouteBuilder: (string -> string -> string) option
+        BaseUri: Uri
+        Client: HttpClient option
+        AuthorizationToken: string option
+        IsBinarySerialization: bool
+        CustomHeaders: (string * string) list
     }
 
     // Warning: if routeBuilder returns a path with a leading /, it will override any subpath inside baseUrl
-    let createApi (baseUrl: string) (routeBuilder: string -> string -> string) =
+    let createApi (baseUrl: string) =
         {
-            routeBuilder = routeBuilder
-            baseUri = Uri(baseUrl)
-            client = None
-            authToken = None
-            isBinarySerialization = false
-            customHeaders = []
+            RouteBuilder = None
+            BaseUri = Uri(baseUrl)
+            Client = None
+            AuthorizationToken = None
+            IsBinarySerialization = false
+            CustomHeaders = []
         }
 
-    let withAuthorizationHeader token options = { options with authToken = Some token }
+    let withRouteBuilder (routeBuilder: string -> string -> string) options = { options with RouteBuilder = Some routeBuilder }
 
-    let withBinarySerialization options = { options with isBinarySerialization = true }
+    let withAuthorizationHeader token options = { options with AuthorizationToken = Some token }
 
-    let withHttpClient client options = { options with client = Some client }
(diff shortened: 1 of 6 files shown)


Fable.Remoting.DotnetClient/ClientRemoting.fs
```diff
@@ -0,0 +1,174 @@
...
+                                Some (args.[0], Some args.[1])
+                            elif generic = typedefof<Async<_>> then
+                                Some (t.GetGenericArguments().[0], None)
+                            else
+                                failwithf "Bad API record field %s, must be of type Async<'a> or a function returning Async<'a>" param.Name
+                        | None -> None
+                    )
+                    |> Seq.toArray
+
+                let route = Uri(options.BaseUri, routeBuilder t.Name param.Name) |> string
```
Zaid-Ajaj, line 152: > Warning: if routeBuilder returns a path with a leading /, it will override any subpath inside baseUrl

Maybe we can have some flexible string handling here to allow the `BaseUri` to end with `/` and the output of the `routeBuilder` to start with `/` with problems. I believe it goes something like this:
```fs
let baseUri = string options.BaseUri
let protocolRoute = routeBuilder t.Name param.Name
let route = baseUri.TrimEnd('/') + protocolRoute.TrimStart('/')
```
To be honest, not sure how `param.Name` becomes the function name of the protocol but I will have a deeper dive into it to understand the implementation

Zaid-Ajaj (approved):
Hi @Arshia001, 

Thanks a lot for giving this a go, I really like this implementation 😍 as it much more flexible than with what quotations provided. The code details look good, I only had one minor remark regarding the concatenation of the base url and the route builder but it should be something really small. 

Give me a sign to let me know if the PR is ready to go on your side of things so that I could publish the goodies 🚀 

There is also the part to update the docs but I can do that once the PR is merged in :smile:

Fable.Remoting.DotnetClient/ClientRemoting.fs
Arshia001: Agreed, most times when a path starts with a leading /, it's not because someone wanted to override the subpath inside the base URL, but I like `Uri`'s handling of edge cases and errors. I think it'd be appropriate if we just strip the leading / from the route part.

> To be honest, not sure how param.Name becomes the function name of the protocol

F# compiles record types to sealed classes containing a single constructor, the names of the parameters to which are the names of the fields. So if you have `type MyType = { x: int; y: int }` the constructor would be `public MyType(int x, int y)`. Unless the record is marked as `CLIMutable`, which is an edge case I'll work around in the next commit.

Arshia001:
@Zaid-Ajaj I just fixed a couple of edge cases (including the leading slash thing) and added documentation. I think the code is good to go. Care to take one more look?

Zaid-Ajaj:
@Arshia001 Awesome work! One last thing I wanted to do which is renaming `ClientRemoting` to just `Remoting` to keep e everything consistent with how the Fable client looks like.  What do you think? If that's OK I will merge and publish right away :smile:

Arshia001:
I did that initially. It's ambiguous with the server-side `Remoting` and I got a ton of errors, so I decided against it. I'm OK with either choice.

Zaid-Ajaj:
Then I will give it a try to rename the module later tonight after finishing some work. I will let you know as soon as I've published the package to nuget :smile:

Again, thanks a lot for the PR, this is definitely much much better than the previous implementation ❤️ 🙏

Arshia001:
I mostly believe I should be the one thanking you for this awesome library, but you're welcome ☺️

Zaid-Ajaj:
Publish the dotnet client as of v3.9 :rocket: 🎉 Added some documentation to the builder functions, renamed `ClientRemoting` to just `Remoting`, doubled the tests of the client with both the quotation proxy and this proxy. 

Please take it for a spin and let me know how it goes :smile:

Arshia001:
I probably should have done all of that myself, sorry about that. I'm actually hoping to use this as the basis for type-safe communication between micro services. I'll take it for a spin tomorrow and let you know.

Zaid-Ajaj:
> I probably should have done all of that myself, sorry about that. 

Oh don't say that, you already did the hard parts, I just threw some cosmetics on top :smile: Good luck with the microservices, I will hear it from you if you encounter any issues :wink:

Arshia001:
Looking absolutely gorgeous 😍

tforkmann:
I'm getting a weird error from Fable:

`ERROR in ./.fable/Fable.Remoting.Client.6.12.0/Remoting.fs
Module build failed (from ./node_modules/fable-loader/index.js):
Error: error recovery at C:/Users/tforkmann/Documents/1_Tests/PIvy/.fable/Fable.Remoting.Client.6.12.0/Remoting.fs (42,10--42,26) IsSynthetic=false (C:/Users/tforkmann/Documents/1_Tests/PIvy/.fable/Fable.Remoting.Client.6.12.0/Remoting.fs)
    at C:\Users\tforkmann\Documents\1_Tests\PIvy\node_modules\fable-loader\index.js:98:22
    at processTicksAndRejections (internal/process/task_queues.js:85:5)
 @ ./src/Client/FetchData.fs 35:0-413 73:16-54 74:15-60 75:13-65 76:9-61 85:19-57 86:15-60 87:13-65 88:9-61 97:19-57 98:15-60 99:13-65 100:9-61 109:19-57 110:15-60 111:13-65 112:9-61 121:19-57 122:15-60 123:13-65 124:9-61 133:20-58 134:16-61 135:13-65 136:9-61 145:20-58 146:16-61 147:13-65 148:9-61 157:20-58 158:16-61 159:13-65 160:9-61 169:20-58 170:16-61 171:13-65 172:9-61 181:20-58 182:16-61 183:14-66 184:9-61 193:20-58 194:16-61 195:14-66 196:9-61
 @ ./src/Client/Client.fs
 @ ./src/Client/Client.fsproj
 @ multi ./src/Client/Client.fsproj`

Any idea why?

Shmew:
Nothing at all should have changed regarding the Fable implementation. Obligatory "have you tried deleting`.fable`"?

tforkmann:
Hmm I did deleted it. Maybe it has to do with Fable itself. Fable.SimpleJson got reccently updated.

tforkmann:
Yes. For sure it has to do something with Fable.SimpleJson. Only upgraded the server part. That works now. Just can't upgrade Client side yet.

Zaid-Ajaj:
@tforkmann Latest Fable.Remoting.Client is at 6.12.0 right now, to which version you want to upgrade exactly?

Shmew:
I'm seeing other reports of this exact issue. Might it be a good idea to un-list 3.12 and republish that as a prerelease version?

Zaid-Ajaj:
I really thought I published a version with a non-breaking changes 😞 I will see if can publish another one on top of it that reverts the breaking stuff. It has to do with the serialization aspect of the SimpleJson: now it is using reflection and a generic type info as argument to become compatible with Fable 2.x and Fable 3 but it seems like existing libraries can't upgrade just yet because of the generic signature of `Json.stringify`. 

Also, yes, SimpleJson is already Fable 3-compatible with all the tests running on node.js and the headless browser.

Zaid-Ajaj:
@tforkmann Can you trying upgrading again, now there should be a new version of Fable.SimpleJson (v3.14) which should be backward-compatible with the rest of the libraries. Give it a try and let me know how it goes, if the problem persists, please open an issue to discuss it further.

PR #187, opened 2020-10-08
Fix endpoint name discovery in the Giraffe adapter

Arshia001:
This change allows Giraffe subroutes (`subRoute "/api" >=> ...`) and, by extension, Saturn's `forward` (`router { forward "/api" ... }`) to be used with `Remoting.buildHttpHandler`.


Zaid-Ajaj:
Thanks a lot @Arshia001 for the fix, I never thought of fixing this issue :smile: A new version of the Giraffe adapter should be up and running as of v4.11

PR #189, opened 2020-10-10
Use FSharpType and FSharpValue instead of direct use of reflection

Arshia001:
OK, I **just** learned about `FSharpType` and `FSharpValue`. I updated the remoting code in the dotnet client to work with `FSharpType`. Functionally, the code does the exact same thing. It's just more readable now. Sorry about this, really. 🙇🏻‍♂️

### DIFF
--- a/Fable.Remoting.DotnetClient/Remoting.fs
+++ b/Fable.Remoting.DotnetClient/Remoting.fs
@@ -4,6 +4,7 @@ open FSharp.Core.OptimizedClosures
 open System
 open System.Net.Http
 open System.Reflection
+open Microsoft.FSharp.Reflection
 
 module Remoting =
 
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
> It's just more readable now. Sorry about this, really. 🙇🏻‍♂️

No worries at all @Arshia001 :smile: I noticed it in the original PR but didn't mind it since it worked perfectly. This makes it more readable indeed, thanks a lot :pray:

Arshia001:
Next time, tell me before I embarrass myself please 😅

Zaid-Ajaj:
Will do :wink:

Arshia001:
@Zaid-Ajaj Say, remember how I said this code was exactly equivalent to the one before? Turns out it's not.

See, F# generates a class constructor for each record. Parameters to that constructor are named similarly to the record's fields, except that they're converted to snakeCase if the field name is PascalCase. This means that the previous code generates a snakeCased version of the route: `MyApi/someHandler` instead of `MyApi/SomeHandler`.

Since Giraffe is normally case-sensitive and FSLint forces record field names to be PascalCase, this makes the previous code essentially unusable. This PR already fixes the error, and it's already been merged. Just one package update and all would be in order.

Again, I'm really sorry about all of this.

Zaid-Ajaj:
Hi @Arshia001, don't worry about it, I just published version 3.10 of Fable.Remoting.DotnetClient that uses this implementation and it should be available in a moment :smile: please give it a try and let me know if I broke something in the process

Arshia001:
Everything seems to be in order. Thanks :)

PR #193, opened 2020-10-28
Do not instantiate API record until it is necessary

kerams:
Implements the second point in https://github.com/Zaid-Ajaj/Fable.Remoting/pull/177 all the way.

It's just a tiny optimization, so there's no need to release it on its own.

### DIFF
--- a/Fable.Remoting.AspNetCore/Middleware.fs
+++ b/Fable.Remoting.AspNetCore/Middleware.fs
@@ -93,8 +93,7 @@ module internal Middleware =
         
         fun (next: HttpFunc) (ctx: HttpContext) -> task {
             let isProxyHeaderPresent = ctx.Request.Headers.ContainsKey "x-remoting-proxy"
-            let impl = implBuilder ctx
-            let props = { Implementation = impl; EndpointName = ctx.Request.Path.Value; Input = ctx.Request.Body; IsProxyHeaderPresent = isProxyHeaderPresent;
+            let props = { ImplementationBuilder = (fun () -> implBuilder ctx); EndpointName = ctx.Request.Path.Value; Input = ctx.Request.Body; IsProxyHeaderPresent = isProxyHeaderPresent;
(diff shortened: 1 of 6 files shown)


Zaid-Ajaj:
Hi there @kerams,

Can you please tell where the optimization is done here? The PR only delays seems building of the implementation for later in the pipeline but it is already delayed since it depends on the incoming request and the execution of the awaitable task 🤔 am I missing something, maybe?

kerams:
The record is created before the current handler is confirmed to be a match. `props` below contains an implementation even though it might not be used if the requested function is not present.
 
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/3ebe8eee62ba981b7ea43e71ce971605905ea1a6/Fable.Remoting.Server/Proxy.fs#L154-L157

This is especially noticeable when you have lots of functions split into multiple separate APIs/records

```fsharp
 .UseGiraffe (choose [ api1Handler; api2Handler; api3Handler, api4Handler; api5Handler; api6Handler ])
```

An incoming request for a function in the last handler currently causes the instantiation of implementations for APIs 1 through 5.

Zaid-Ajaj:
Thanks for the explanation, that makes sense. Though I had in mind that those instantiations are really cheap to begin with but it doesn't hurt to push for more optimizations 🚀

kerams:
Sure, though they can potentially instantiate more expensive dependencies of their own. If using the approach described in the docs

```fsharp
let createTodoApiFromContext (httpContext: HttpContext) : ITodoApi = 
    let todoStore = httpContext.GetService<ITodoStore>()
    createTodoApi todoStore
```

Zaid-Ajaj:
Yeah it is good not to assume or expects things from user code (like always having a cheap factory function)

Zaid-Ajaj:
Published new versions of Server (base) and adapters Suave, Giraffe and AspNetCore which include the improvements in this PR 🚀

PR #196, opened 2020-11-15
Fix array of tuples deserialization in Fable

kerams:
A workaround for a bug in Fable (https://github.com/fable-compiler/Fable/pull/2272).


Zaid-Ajaj:
Published MsgPack v1.6.2 and Client v7.2 packages which include this fix 🚀 🚀

PR #203, opened 2020-12-10
minimum code to handle DataTable serialization through BinaryFormatter

smoothdeveloper:


Fable.Remoting.MsgPack/Write.fs
```diff
@@ -228,6 +230,16 @@ let inline writeGuid (g: Guid) out =
 let inline writeBigInteger (i: bigint) out =
     writeBin (i.ToByteArray ()) out
 
+let inline writeDataTable (dt: System.Data.DataTable) out =
+  if String.IsNullOrWhiteSpace dt.TableName then
+    failwithf "datatable must have TableName property set"
+  do
```
kerams, line 236: Why `do`? :)

Fable.Remoting.MsgPack/Write.fs
```diff
@@ -367,8 +379,12 @@ and private makeSerializerAux<'T> (ctx: TypeGenerationContext): Action<'T, Strea
...
         let elementSerializers = shape.Elements |> Array.map makeMemberVisitor
         Action<_, _> (fun (tuple: 'T) out -> writeTuple tuple out elementSerializers) |> w
-    | _ ->
-        failwithf "Cannot serialize %s." typeof<'T>.Name
+    | _shape ->
+        let tableTy =  typeof<System.Data.DataTable>
+        if typeof<'t> = tableTy ||  typeof<'t>.IsInstanceOfType tableTy then
+          Action<_, _> (fun (table: System.Data.DataTable) out -> writeDataTable table out) |> w
+        else
+          failwithf "Cannot serialize %s." typeof<'T>.Name
```
kerams, line 387: There's nothing inherently wrong with this, but I'd rather define a shape recognizer (active pattern) for data tables to keep the big pattern match clean. You can base it on this one:

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/7e3c52003c7390e57a8ea6e6f1e4afac3f12d199/Fable.Remoting.Server/Proxy.fs#L46-L49

kerams (changes requested):
You need this for .NET <-> .NET communication, right? I don't think data tables are supported by Fable, so this would be the first thing to break parity between the Fable and .NET implementations in terms of supported types, not that I really have a problem with that.

Fable.Remoting.MsgPack/Write.fs
Zaid-Ajaj: > if String.IsNullOrWhiteSpace dt.TableName then

Doesn't this sound like an application specific logic and not so much a data transport concern or am I missing something? 🤔

smoothdeveloper: @kerams it is useful to scope the `use` of stuff and signal 'side-effectey' block, but the code was pasted there and I didn't notice in the short body, it doesn't matter.

@Zaid-Ajaj I've removed it, I was actually dealing with having to debug the server side and it was just the intent to help user somehow, but it can be done outside the library, no problem. I've updated the code.

kerams:
The 2 new `!FABLE_COMPILER` directives in Write.fs are not needed as you're already inside of a big block that is conditionally compiled on .NET (lines 12 - 433).

Zaid-Ajaj:
@smoothdeveloper The changes look great! Besides the last comment by @kerams, there is just one more thing: adding an end-to-end test the communication. This should be easy if you look inside `Fable.Remoting.Giraffe.Tests`, add a new remote function (record field) to the `IBinaryServer` protocol like an echo function: `echoDataTable : DataTable -> Async<DataTable>` and test the round trip in the `MiddlewareTests.fs` file 🙏

kerams:
@Zaid-Ajaj, JS <-> .NET integration tests won't work since Fable does not support data tables, right?

Zaid-Ajaj:
> JS <-> .NET integration tests won't work since Fable does not support data tables, right?

@kerams That's right, I think using a `DataTable` type won't even compile in a Fable app

smoothdeveloper:
@Zaid-Ajaj I've adressed the review comments, although adding the test in `MiddlewareTests.fs` gave faulty roundtrip somehow.

@kerams I think I've followed correctly the advice for active pattern, thanks to both for showing me around.

kerams (approved):

Zaid-Ajaj:
@smoothdeveloper I am afraid the reason you had failing tests is because the serialization and deserialization is not entirely correct :/ first of all, the `TableName` isn't going back and forth correctly. This was the original test:
```fs
// input table
t.TableName <- "myname"
// assert
Expect.equal deserialized.TableName  t.TableName  "table name"
```
When I added full roundtrip tests, I though maybe it is just the `TableName` that was incorrect but actually even the types are deserialized incorrectly, an `int` column from the input table became an `int64` when deserialized. I tried a bunch of combinations to ensure exact conversions but no luck. I think this needs more refinement and maybe a more fledged out intermediate binary representation that includes the table name, schema and data separately from which the data table can be reconstructed on the other end. 

I will revert this PR for now, please give it another try on your branch because my attempts have so far failed.

smoothdeveloper:
@Zaid-Ajaj sorry for you encountering the issues! It has been trial and error but I also noticed the type conversions after the roundtrip in the second test.

I'll try to get it working, there are few bits that need to be taken care of.

PR #204, opened 2020-12-10
Update appveyor dotnet SDK version (7th attempt)

kerams:
Oof, that took some doing. I think the paket version warnings are caused by the fact that Fake is still using v5.


Zaid-Ajaj:
Thanks a lot 💯 I just gave up on this morning after idk how many attempts 😓

PR #208, opened 2020-12-12
another try for support of DataTable and DataSet

smoothdeveloper:
It works through WriteXmlSchema and WriteXml, it seems to roundtrip ok this time.

The issue was that newtonsoft.json had a very poor encoding of those objects and that I had no clue of necessary changes to FableConverter to override that.

I haven't looked from javascript client, but it should get two xml strings which can be used to assemble the same data.

### DIFF
--- a/Fable.Remoting.Giraffe.Tests/MiddlewareTests.fs
+++ b/Fable.Remoting.Giraffe.Tests/MiddlewareTests.fs
@@ -575,4 +575,55 @@ let middlewareTests =
             let expected = Map.ofList [ "hello", 5; "there!", 6 ]
             Expect.equal output expected "Echoed map is correct"
         }
+
+        testCaseAsync "IBinaryServer.echoDataTable" <| async {
+            let t = new System.Data.DataTable()
+            t.TableName <- "myname"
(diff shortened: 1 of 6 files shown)


Zaid-Ajaj:
Code looking good @smoothdeveloper! 

I see that you have also added support inside of the JSON converter, can you please add tests for those?
 - Inside Fable.Remoting.Json.Tests
 - Inside Fable.Remoting.Giraffe.Tests within for the `IServer` protocol

> I haven't looked from javascript client, but it should get two xml strings which can be used to assemble the same data.

We don't need to even bother with the client JS because: 
 - Fable won't compile the `DataSet` and `DataTable` 
 - SimpleJson would need a Xml parser which we have but it is a lot of work to incorporate it into the parser

Zaid-Ajaj:
It's probably annoying to keep asking you for changes, I will just try to add them myself and hopefully it all works out 😄

Zaid-Ajaj:
Added tests, everything worked as expected 🎉 so I published all the goodies to nuget. Update your packages to latest and you should have the implementation working for you 🚀

Issue #217, opened 2021-02-11
withCredentials false does not work, still getting CORS related issues related to the fact that the request has credentials


AntonioL:
I am trying to send a request by setting withCredentials false otherwise I will hit CORS related issues.

The same request done via cURL works perfectly meaning that this only fails in browsers.

This is the error: https://developer.mozilla.org/en-US/docs/Web/HTTP/CORS/Errors/CORSNotSupportingCredentials

I am using 7.2.0 version of Fable.Remoting.Client

Reproducible example:
```fsharp
module Client

open btb.core
open Elmish
open Elmish.React
open Fable.React
open Fable.React.Props
open Fable.Remoting.Client



// UPDATE

let api =
    Remoting.createApi()
    |> Remoting.withCredentials false
    |> Remoting.withBaseUrl "http://127.0.0.1:8083/"   
    |> Remoting.buildProxy<IGreetingApi>

.....
.....
async {
  let! num = api.sum 1
  printfn "Next number after 1 is %d" num
}
|> Async.StartImmediate


```

AntonioL:
I worked around the issue on the server side amending the CORS Policy by allowing that origin.

AntonioL:
Closing the issue.

PR #223, opened 2021-03-09
Fix withCredentials = false not working

KittisakT:
Related issue #217

### DIFF
--- a/Fable.Remoting.ClientV2/Http.fs
+++ b/Fable.Remoting.ClientV2/Http.fs
@@ -55,7 +55,7 @@ module Http =
             for (key, value) in req.Headers do
                 xhr.setRequestHeader(key, value)
 
-            xhr.withCredentials <- true
+            xhr.withCredentials <- req.WithCredentials
 
             xhr.onreadystatechange <- fun _ ->
                 match xhr.readyState with


Zaid-Ajaj:
Thanks a lot @KittisakT I don't how I missed that one 😓 just merged and published Fable.Remoting.Client v7.5 that includes the fix 🚀 should be on nuget in a moment 😄

kerams:
`WithCredentials` is initialized to `false` so this also changes the current default behavior.

Zaid-Ajaj:
Should we make initialization optional until the user provides a value? i.e. use `bool option` in the config? According to [this](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials), the default is `false`

kerams:
That could still break something after upgrading for consumers that rely on the current ("incorrect/buggy") behavior. Then again, there's no way around it if we're going to follow the guidance you posted, which I'd be in favor of.

So in the end maybe just keep it as it is? Just felt like pointing out the potential problem 😄.

Zaid-Ajaj:
> Then again, there's no way around it if we're going to follow the guidance you posted, which I'd be in favor of.

I was thinking the same, we can't do much about it, will just keep it as is

PR #228, opened 2021-03-13
Fixes #226 and Fixes #227

### LINKED ISSUE #226: Update TypeShape 8.0.1 -> 9.0.0
Some weird _FATAL_ runtime error occurs with TypeShape and it happens only on deployed application in azure (assuming it is running net5.0) but works locally just fine using either -c Debug or -c Release

```
Fatal error. Internal CLR error. (0x80131506) at System.RuntimeType.get_IsByRefLike() at TypeShape.Core.Core+TypeShapeImpl.resolveTypeShape(System.Type) at TypeShape.Core.Core+ShapeFSharpRecord`1[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]]..ctor() at System.RuntimeMethodHandle.InvokeMethod(System.Object, System.Object[], System.Signature, Boolean, Boolean) at System.Reflection.RuntimeConstructorInfo.Invoke(System.Reflection.BindingFlags, System.Reflection.Binder, System.Object[], System.Globalization.CultureInfo) at TypeShape.Core.Core+TypeShapeImpl.activateGeneric(System.Type, System.Type[], System.Object[]) at TypeShape.Core.Core.Activator.CreateInstanceGeneric.Static[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Microsoft.FSharp.Core.FSharpOption`1<System.Type[]>, Microsoft.FSharp.Core.FSharpOption`1<System.Object[]>) at TypeShape.Core.Core+Shape.|FSharpRecord|_|(TypeShape) at Fable.Remoting.Server.Proxy.makeApiProxy[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e],[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Fable.Remoting.Server.RemotingOptions`2<System.__Canon,System.__Canon>) at Fable.Remoting.Giraffe.GiraffeUtil.buildFromImplementation[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Microsoft.FSharp.Core.FSharpFunc`2<Microsoft.AspNetCore.Http.HttpContext,System.__Canon>, Fable.Remoting.Server.RemotingOptions`2<Microsoft.AspNetCore.Http.HttpContext,System.__Canon>) at Fable.Remoting.Giraffe.Remoting.buildHttpHandler[[System.__Canon, System.Private.CoreLib, Version=5.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e]](Fable.Remoting.Server.RemotingOptions`2<Microsoft.AspNetCore.Http.HttpContext,System.__Canon>) at <StartupCode$Server>.$Server.main@()
```
So I am thinking it might have to do with TypeShape using an old reflection API so I think we should update it to latest 9.0.0 just in case. What do you think, @kerams?

### LINKED ISSUE #227: MsgPack DateTime timezone unsupported
Hi there,

I'm using Fable.Remoting.MsgPack serialization. In the server we specify our DateTime objects with kind UTC. After sending them through Remoting to the client, the timezone kind is no longer UTC.

After checking the implementation I noticed that the transfered ticks are passed into the DateTime contstructor without any timezone kind. If I understand it correct, there is no possibility to transfer timezone specific DateTimes through MsgPack.

Did I missed something or is this the expected behaviour?

Kind regards
Christoph

Zaid-Ajaj:
This PR includes TypeShape source file and removes the problematic code that checks for `IsByRefLike` dynamically since we don't need it afaik (fixes #226) I've also attempted to solve #227 where DateTimeKind wasn't preserved when sent over the wire (also because I wanted to dive into the MsgPack implementation a little bit)  

@kerams Can you please have a look just to make sure I didn't do anything too dumb? Besides that I used the space of a `int64` to send a enum which can only have values {0, 1, 2} 😅

### DIFF
--- a/Fable.Remoting.MsgPack/Read.fs
+++ b/Fable.Remoting.MsgPack/Read.fs
@@ -25,7 +25,6 @@ let inline interpretIntegerAs typ n =
     if typ = typeof<Int32> then int32 n |> box
     elif typ = typeof<Int64> then int64 n |> box
     elif typ = typeof<Int16> then int16 n |> box
-    elif typ = typeof<DateTime> then DateTime (int64 n) |> box
     elif typ = typeof<UInt32> then uint32 n |> box
     elif typ = typeof<UInt64> then uint64 n |> box
     elif typ = typeof<UInt16> then uint16 n |> box
@@ -114,10 +113,10 @@ type Reader (data: byte[]) =
 #else
         if BitConverter.IsLittleEndian then
             let arr = Array.zeroCreate len
-        
+
             for i in 0 .. len - 1 do
                 arr.[i] <- data.[pos + len - 1 - i]
-            
+
             pos <- pos + len
             m (arr, 0)
         else
@@ -171,24 +170,24 @@ type Reader (data: byte[]) =
 #if !FABLE_COMPILER
         mapReaderCache.GetOrAdd (t, Func<_, _>(fun (t: Type) ->
             let args = t.GetGenericArguments ()
-            
+
             if args.Length <> 2 then
                 failwithf "Expecting %s, but the data contains a map." t.Name
 
             let mapDeserializer = typedefof<DictionaryDeserializer<_,_>>.MakeGenericType args
             let isDictionary = t.GetGenericTypeDefinition () = typedefof<Dictionary<_, _>>
             let d = Delegate.CreateDelegate (typeof<Func<int, bool, (Type -> obj), obj>>, mapDeserializer.GetMethod "Deserialize") :?> Func<int, bool, (Type -> obj), obj>
-            
+
             fun (len, x: Reader) -> d.Invoke (len, isDictionary, x.Read))) (len, x)
 #else
         let args = t.GetGenericArguments ()
-        
+
         if args.Length <> 2 then
             failwithf "Expecting %s, but the data contains a map." t.Name
 
         let pairs =
             let arr = Array.zeroCreate len
-            
+
             for i in 0 .. len - 1 do
                 arr.[i] <- x.Read args.[0] |> box :?> IStructuralComparable, x.Read args.[1]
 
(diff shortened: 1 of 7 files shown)


Fable.Remoting.MsgPack/Read.fs
```diff
@@ -335,24 +334,34 @@ type Reader (data: byte[]) =
...
             FSharpValue.MakeTuple (FSharpType.GetTupleElements t |> Array.map x.Read, t)
 #endif
+        elif t = typeof<DateTime> then
+            let dateTimeTicks = x.Read typeof<int64> :?> int64
+            let kindAsInt = x.Read typeof<int64> :?> int64
+            let kind =
+                match kindAsInt with
+                | 1L -> DateTimeKind.Utc
+                | 2L -> DateTimeKind.Local
+                | _ -> DateTimeKind.Unspecified
```
kerams, line 344: Enums are supported so you can simply do
```fsharp
let kind = x.Read typeof<DateTimeKind> :?> DateTimeKind
```
and on the .NET write side (might have to move `writeEnum` higher up)
```fsharp
writeEnum dt.Kind out
```

It might be ever so slightly slower than what you did with explicit integer to enum mapping. Up to you ;)

kerams (approved):
There's no problem with using `writeInt64`. It won't write more bytes than necessary for each individual integer (could be 1, 2, 3, 5, or 9 bytes). In this case 1 byte, because all `DateTimeKind` values are in the range of [fixnum](https://github.com/msgpack/msgpack/blob/master/spec.md#int-format-family).

Or you could also choose to directly write the enum as fixnum for extra speed using `out.WriteByte (Format.fixposnum (int dt.Kind))` since we can be sure of this enum's range. I used it for union tags

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/c65f6040a7e74eac471e22015ee133fd8b19c986/Fable.Remoting.MsgPack/Write.fs#L254-L259

Zaid-Ajaj:
Thanks a lot for the explanation @kerams if the way I handled DateTimeKind isn't problematic and works correctly, I will leave it as is. Let's get this in and publish a new batch 😄

kerams:
Unfortunately, MsgPack also depends on TypeShape, so it'll need to be included there too.

PR #236, opened 2021-04-04
Fix MsgPack decimal

### LINKED ISSUE #235: MsgPack invalid casting
I'm encountering an exception when initializing the .NET MsgPack serializer. I tried finding where this casting is happening, but no luck. The exception on my end is happening [here](https://github.com/Shmew/Fable.SignalR/blob/master/src/Fable.SignalR.Shared/MsgPack.fs#L32). I'm having trouble properly reproducing the exception as it isn't happening in my actual repo (I tried both updating and downgrading packages with no luck), but is for [others](https://github.com/Shmew/Fable.SignalR/issues/21). SignalR swallows the exception normally so the reproduction he linked won't reveal it, I had to reference the `Fable.SignalR` projects directly and run the debugger to find this:

image

If I can do anything else to help please let me know.

kerams:
Decimals were be serialized as floats, which could result in losing precision (or rather the deserialized number would not match the original). Instead, we now use its binary representation using 4 integers.

Also fixes #235 with some better messages.

### DIFF
--- a/ClientV2Tests/src/App.fs
+++ b/ClientV2Tests/src/App.fs
@@ -1239,7 +1239,7 @@ let binaryServerTests =
                 let! output = binaryServer.echoInt64WithMeasure input
                 test.equal true (input = output)
 
-                let input = 333.35M<SomeUnit>
+                let input = 32313213121.1415926535m<SomeUnit>
                 let! output = binaryServer.echoDecimalWithMeasure input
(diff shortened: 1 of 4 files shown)


Zaid-Ajaj:
Awesome stuff :fire: Published as of v1.9 for Fable.Remoting.MsgPack

PR #238, opened 2021-04-19
Add cancellation support

do-wa:
Hi,

I've created a basic implementation of cancellation support by leveraging the f# async implicit token passing. I think the unit test shows a simple application for this. 

I'm making this a draft since I'm not sure if the integration is missing some key points or violates some guidelines for this project. 

Hope this helps.

PR #239, opened 2021-04-19
Consolidate `send` and `sendAndReadBinary` implementations into one method

schauerte:
We came across this somewhat duplicated code (of `send` and `sendAndReadAsBinary`) while trying to add cancellation support to Fable.Remoting. So to not end up changing both of these methods for this, we would kindly suggest using only one implementation for handling the `XMLHttpRequest`.

I hope this request is considered meaningful and our changes are a feasible way to achieve this - otherwise please let us know.

### DIFF
--- a/Fable.Remoting.ClientV2/Http.fs
+++ b/Fable.Remoting.ClientV2/Http.fs
@@ -42,8 +42,7 @@ module Http =
     /// Appends a request with string body content
     let withBody body (req: HttpRequest) = { req with RequestBody = body }
 
-    /// Sends the request to the server and asynchronously returns a response
-    let send (req: HttpRequest) =
+    let private sendAndRead (preparation:(XMLHttpRequest -> unit) option) resultMapper (req: HttpRequest) =
         Async.FromContinuations <| fun (resolve, _, _) ->
             let xhr = XMLHttpRequest.Create()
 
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @schauerte to be honest with you, I don't mind code duplication especially in cases where the code is pretty simple. I am wondering though, is this PR part of #238 (add cancellation support) and if so, why not make into the same PR?

schauerte:
#238 is exactly, why we came across this - at first we only enhanced one of these methods (overlooked the other one) and wondered why our changes do not work...
Of course duplication (of simple code) is fine - we just differ in what looks simple to us.

I wanted to clarify *if* you are OK with the consolidation or prefer the separated approach (for any reason) before creating a pull request for the actual cancellation change - but @do-wa and I talked past each other and he went along creating #238. We are happy to create one PR for both . just let us know if we should combine them or simply abandon this one.

Zaid-Ajaj:
>  let us know if we should combine them or simply abandon this one.

Would be great if you can combine into one PR please 🙏 thanks!

schauerte:
moved to #238: https://github.com/Zaid-Ajaj/Fable.Remoting/pull/238/commits/d498bf75c0825f57cf34c38ef1aa486207cf3b49
Thank you very much and sorry for the confusion.

PR #240, opened 2021-04-22
Expose the request body to the error handler

kerams:
While this is technically a breaking change, I reckon it's worth adding. It allows the user to log the entire request body (if it was JSON), which is not only helpful for diagnosing general application errors, but especially so in cases when the input is malformed and not deserializable internally by Fable.Remoting (for whatever reason).

### DIFF
--- a/Fable.Remoting.AspNetCore/Middleware.fs
+++ b/Fable.Remoting.AspNetCore/Middleware.fs
@@ -76,14 +76,14 @@ module internal Middleware =
       fun (next : HttpFunc) (ctx : HttpContext) ->
         task { return None }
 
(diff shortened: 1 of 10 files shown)


kerams:
Not a breaking change anymore :).

Zaid-Ajaj:
Hi @kerams, this looks like a nice addition to the error handler since it has access to the context (unlike the [diagnosticsLogger](https://zaid-ajaj.github.io/Fable.Remoting/src/troubleshoot.html)) 

I was just going to ask you to add the property to the RouteInfo but you already did it 😍 

I will merge it and publish ASAP

Zaid-Ajaj:
Published the packages 🚀 you should be able to update to latest to get the goodies

PR #246, opened 2021-05-15
Refactor JS client proxy

kerams:
Cleaned up unused code, merged almost duplicate branches, and moved some static data out of `fun arg0 arg1 arg2 arg3 arg4 arg5 arg6 arg7 ->` so that it isn't recomputed on every request.

### DIFF
--- a/Fable.Remoting.ClientV2/Proxy.fs
+++ b/Fable.Remoting.ClientV2/Proxy.fs
@@ -22,33 +22,10 @@ module internal Blob =
             reader.readAsText(blob)
 
 module Proxy =
-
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Hi @kerams this one looks good, do you know if it resolve #249? 'Cause I don't see checking for a specific status code of zero

kerams:
No, that stuff is in the Http module. This PR should not include any functional changes.

Zaid-Ajaj:
Working on a fix that removes the check `status code <> 0`  the Http module and will push out a new version.

PR #252, opened 2021-06-03
Internalize TypeShape, take #2

kerams:


kerams:
Any ideas about excluding the two files from the Nuget package?

Zaid-Ajaj:
> Any ideas about excluding the two files from the Nuget package?

You can conditionally `Compile` the files based on the `FABLE_COMPILER` I believe so that Fable won't compile them but I guess compiler directives like `#if !FABLE_COMPILER` in the files themselves are better because then the tests would catch any funky business that is otherwise not visible until someone compiles the nuget and it breaks

kerams:
Added. The files will still be shipped, but I guess it's no biggie.

Zaid-Ajaj:
A fresh batch is published to include this fix, let's see if the SAFE template issue with 32-bit apps is fixed 🤞

PR #255, opened 2021-06-11
fix int64 serialization

christoph-feb22:
Hi Zaid,

I noticed a bug on realy rare cases. On serializing int64 to MsgPack, while there are bytes with value 0 at position 5 or 6 in the byte representation, these positions will be removed. This leads to an invalid case. There is an int64 number with to less byte positions.

The byte representation of the number in the test case is: 8 217 43 82 0 0 153 102.
In the old implementation the 0 will be removed.

### DIFF
--- a/ClientV2Tests/src/App.fs
+++ b/ClientV2Tests/src/App.fs
@@ -333,8 +333,10 @@ let serverTests =
             async {
                 let! fstResult = server.echoPrimitiveLong System.Int64.MaxValue
                 let! sndResult = server.echoPrimitiveLong System.Int64.MinValue
+                let! thirdResult = server.echoPrimitiveLong 637588453436987750L
                 do test.equal true (fstResult = System.Int64.MaxValue)
                 do test.equal true (sndResult = System.Int64.MinValue)
+                do test.equal true (thirdResult = 637588453436987750L)
             }
 
         testCaseAsync "IServer.echoComplexLong" <|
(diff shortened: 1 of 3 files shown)


Zaid-Ajaj:
Hi @christoph-feb22 thanks a lot for this the fix, to be honest I don't understand the issue (@kerams is the MsgPack expert) but the tests pass with flying green colors, so I am happy to merge and publish as soon as possible

kerams (changes requested):
This is a great catch. Would you be so kind an also fix this in the Fable part, which suffers from the same issue?

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/f8157ce80cf77770e6f8216f5194afa9bd49a709/Fable.Remoting.MsgPack/Write.fs#L489-L498

Then you can (almost) copy-paste your test case under this one

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/f8157ce80cf77770e6f8216f5194afa9bd49a709/ClientV2Tests/src/App.fs#L1551-L1552

Finally, it would be great to have this covered in an end-to-end test (serialized as JS in Fable, deserialized on the server, serialized with MsgPack on the server and deserialized as MsgPack in Fable). It should only take 2 more lines inside this test

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/f8157ce80cf77770e6f8216f5194afa9bd49a709/ClientV2Tests/src/App.fs#L332-L338

kerams (approved):
Fantastic, thanks.

Zaid-Ajaj:
PR merged and published into Fable.Remoting.MsgPack v1.12 and all downstream packages that depend on it (basically all of them) thanks a lot @christoph-feb22 for the fix 🙏 ❤️

PR #256, opened 2021-06-13
Make MsgPack treeshakable to reduce bundle size when MsgPack isn't used

bklop:
This small change moves the `MsgPack.Read` reference to `Remoting.withBinarySerialization`. As long as you don't call that function, bundlers are free to treeshake the MsgPack stuff out. Previously the hard reference to `MsgPack.Read` in `Proxy.executeRequest` prevented this.

I've tested it on an application using Fable 3.2.2 and Webpack 5.38.1. The production bundle size decreased with 15KB (post-minification, but before compression). Enough to make it worthwhile IMHO 😄.

This shouldn't be a breaking change either, unless someone uses the RemoteBuilderOptions record directly...

### DIFF
--- a/Fable.Remoting.ClientV2/Proxy.fs
+++ b/Fable.Remoting.ClientV2/Proxy.fs
@@ -4,7 +4,6 @@ open System
 open Fable.Core
 open Fable.SimpleJson
 open Browser.Types
-open Fable.Remoting
 
 module internal Blob =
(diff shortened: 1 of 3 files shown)


Zaid-Ajaj:
Published as of Fable.Remoting.Client v7.15 🚀 keep 'em coming 😄

PR #258, opened 2021-06-14
Fix byte serialization in MsgPack

### LINKED ISSUE #257: Error in BinarySerialization
Hi there,

i get following issue with the BinarySerialization:

exn Error: Position 441, byte 201, expected type Byte.

The error is probably coming from this patter match.

https://github.com/Zaid-Ajaj/Fable.Remoting/blob/f04255c7cb8c8c3316d1730c12c1a570975281cb/Fable.Remoting.MsgPack/Read.fs#L441

Any idea how to fix that?

When I get rid of BinarySerialization Fable.Remoting works well.

Thanks in advance!

kerams:
Fixes #257.

@Zaid-Ajaj, I've discovered a problem with JSON `SByte` serialization though. The echo response is `{"Byte":200,"Maybes":[{"Just":-120},"Nothing",{"Just":120},{"Just":5},{"Just":-5}],"SByte":-10}`, which looks correct, but deserialization fails with `Cannot convert -120 to System.SByte`.

```fsharp
type OtherDataC = {
    Byte: byte
    Maybes: Maybe<sbyte> list
    SByte: sbyte
}

let input = {
    Byte = 200uy
    SByte = -10y
    Maybes = [ Just -120y; Nothing; Just 120y; Just 5y; Just -5y ]
}
```

Could be something in SimpleJson?


Zaid-Ajaj:
The problem was indeed in SimpleJson. Just fixed it and published a new package v3.21 which should support sbyte types. 

Can you try updating SimpleJson here when it becomes available? Then I can merge this and publish the goodies :smile:

kerams:
Good stuff, we have liftoff.

Zaid-Ajaj:
Published a new batch 🚀

PR #260, opened 2021-06-23
Update error-handling.md

isaacabraham:
Remove explicit binding - this code gives a warning for unhandled cases, and is unnecessary as we've already bound to Exception in the function argument.


Zaid-Ajaj:
Thanks!

Issue #262, opened 2021-08-01
When publishing packages please enable/support Source Link


s6o:
I'm a newbie when comes .NET and F# ecosystems and I've just been playing around with various things.

An annoying discovery compared to Flutter/Dart stack is that on many F# FOSS packages it not possible to look at the source directly from the IDE.

E.g. was playing around with https://github.com/dsyme/guido-learns-fsharp in VSCode and wanted to "drill down" to actual Remoting.* source, but no luck.

The good news seems to be that this behavior is actually supported in the .NET ecosystem, but would need to be enabled when publishing packages. References:
* https://youtu.be/i3qEhwcG7ps?t=11568 (feature demo in VSCode, .NET Conf: Focus on F#)
* https://docs.microsoft.com/en-us/dotnet/standard/library-guidance/sourcelink
* https://devblogs.microsoft.com/dotnet/producing-packages-with-source-link/

The (obvious) benefits of Source Link enabled packages:
* the code is the basic/fundamental documentation
* improved discovery and learning

Zaid-Ajaj:
Hi there @s6o thanks for reporting the issue! Adding SourceLink support sounds like a nice addition to remoting, I will look into it very soon

PR #270, opened 2021-09-12
[experiment] add SourceLink to the Fable.Remoting.Json nuget package

Numpsy:
I saw #262 and thought that as I've been having a go with Remoting recently and SourceLink might be useful, Id have a go at doing the change.

It's labelled 'experiment' because I've never done it on a project using Paket before - I gather from reading elsewhere that the 'copy_local: true' is the way to make it attach SourceLink to the build, but to *not* make the produced Remoting package have a runtime dependency on the SourceLink packages. So - I wanted to confirm that is ok before doing any of the other packages.

This is what the before (the current package published to nuget.org on the left) and after (a local package build on the right):

image

So the repository information is there.

Note: This is using the SourceLink approach of putting the pdb files in the nupkg - an alternative is to make a snupkg file, but I haven't actually done that before, and don't know what the publishing procedure is.

Related note: There are warnings in the package build about ```PackageLicenseUrl``` being deprecated - any thoughts or objections for a PR to replace it with ```PackageLicenseExpression```?


Numpsy:
On a semi related note, I tried to get my local build marked as deterministic as well, but didn't get it to work (that may or may not be down to https://github.com/dotnet/fsharp/issues/11920, my knowledge of the tools is a bit too thin to be certain)

Zaid-Ajaj:
Since I don't really know how to test this and you are not sure it will work correctly, I will just close it for the time being.

PR #272, opened 2021-09-15
Support TimeOnly, DateOnly, update to .NET 6

kerams:
I've also bumped the `net45` TFM to `net462` since [it is going out of support](https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/) and cannot even be compiled against in VS 2022 anymore. Fable.Remoting.Server is also on `net462`, so this should be fine.


kerams:
Oh, yeah, this is also blocked by the fact that FAKE is for all intents and purposes dead https://github.com/fsharp/FAKE/pull/2587.

kerams:
This will be ready with https://github.com/Zaid-Ajaj/Fable.SimpleJson/pull/82.

Zaid-Ajaj:
Hi @kerams thanks a lot for the PR! I'll try today to properly review and get everything in ✅

paket.dependencies
```diff
@@ -17,7 +17,7 @@ nuget Giraffe
 
 group Build
     source https://api.nuget.org/v3/index.json
-    framework: netstandard2.0
+    framework: net6.0
```
kerams, line 20: Had to change this and remove asyncs from build.fsx, otherwise Fake just refuses to work. No idea what to do about it.

Zaid-Ajaj: I will be removing the `build.fsx` at some point and start using a simple console app for build automation

Zaid-Ajaj:
Published a batch of packages 🚀 🚀 🚀 thanks a lot @kerams 🙏

PR #284, opened 2021-12-02
Make readInt() 3-4 times faster

juselius:
`readint ()` uses a local temporary array for endian conversion. This array gets created and garbage collected for every value, putting a lot of strain on the garbage collector. By making the array a reusable class variable, we achieve 3-4 times faster execution. The class variable is only used by Fable, and since JavaScript is single-threaded there should not be any issues related to race conditions etc.

### DIFF
--- a/Fable.Remoting.MsgPack/Read.fs
+++ b/Fable.Remoting.MsgPack/Read.fs
@@ -94,6 +94,7 @@ type SetDeserializer<'a when 'a : comparison> () =
 
 type Reader (data: byte[]) =
     let mutable pos = 0
+    let intBuf = Array.zeroCreate 8
 
 #if !FABLE_COMPILER
     static let arrayReaderCache = ConcurrentDictionary<Type, (int * Reader) -> obj> ()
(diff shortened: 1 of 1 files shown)


kerams:
Ah, you found perhaps the lowest hanging fruit it would seem :). Very nice.

kerams (approved):

kerams:
You know, on second thought it might good to do the same thing on .NET too. The in-place reverse with `Array.Reverse` is dirty because it mutates the input array if the system is little endian. I would not worry about race conditions on .NET because the `Reader` type is not designed to be thread-safe (`pos` is already a mutable instance property), though it would be problematic should be ever use parallelization during a single deserialization.

What do you think?

juselius:
@kerams since multi-threading and parallelization isn't a concern, it's a trivial fix. Just remove the special case for `!FABLE_COMPILER`, and it's done :) I'll make a new PR!

juselius:
Shall I bump the version to 1.13.1 at the same time?

kerams:
> Shall I bump the version to 1.13.1 at the same time?
There's no need.

Zaid-Ajaj:
Thanks a lot @juselius and @kerams I will publish a new package soon!

Zaid-Ajaj:
@juselius A new batch of packages has been publishes that includes the improvement in MsgPack 🚀 updating `Fable.Remoting.Client` should give you latest bits

Issue #134, opened 2019-09-04
Support standard Tasks


isaacabraham:
Any plans to support regular .NET tasks in addition to `async` workflows?

Zaid-Ajaj:
Hello @isaacabraham, AFAIK Fable doesn't support Task yet, so it doesn't really make sense to support it yet out of the box and the workaround is as simple as `Async.AwaitTask`

isaacabraham:
Yes true. I was thinking more in terms of server-side (i.e. using Task there) but of course that's not possible as we have a typed contract between both sides (which is the whole point of remoting...). Closing :-)

tforkmann:
Just wanted to open the same issue. The workaround is easy but it would be nice to just use tasks instead of async...

forki:
@Zaid-Ajaj could you please elaborate on this? I personally would prefer to use promise {} on client and task {} on giraffe. Is this not something that would make sense?

Zaid-Ajaj:
@forki The types are shared between client and server, this means both have to understand and know how to work with the shared types. In case of task, the client doesn't understand it as of now and in case of promise, the server won't understand it either. How would you write a int -> Promise<string> on the from giraffe.

Async however, is something both client and server know how to work with directly. I'd not implement promises on the client either because of their weird error handling model, they can behave unexpectedly because they are like "hot" tasks, executed when created directly.

forki:
Ok so the whole computation is cross compiled? I thought it would just
cross compile the data transfer type and the computation wrapper would be
platform dependent

Zaid Ajaj <notifications@github.com> schrieb am Sa., 6. Juni 2020, 13:04:

Zaid-Ajaj:
No not the whole computation, just the types are cross-compiled. Async type is understood by both client and server

isaacabraham:
In other words someone would have to let Fable know what task { } should be rendered out into on the JS side.

forki:
Well I don't really understand why the client would want to know if the
internal of the server request is using async or task or returning it sync.

Isaac Abraham <notifications@github.com> schrieb am Sa., 6. Juni 2020,
17:52:

Zaid-Ajaj:
> In other words someone would have to let Fable know what task { } should be rendered out into on the JS side.

@isaacabraham Yes, but first it has to be supported in F# as a built-in computation expression which is now being implemented. 

> Well I don't really understand why the client would want to know if the
internal of the server request is using async or task or returning it sync.

[...]

PR #293, opened 2022-02-13
Support tasks

kerams:
Implements #134.

@isaacabraham, what's the goal here? Did you just want to avoid mucking around with Task <-> Async conversions in your code? This PR delivers on that front (though the directives I mentioned in https://github.com/Zaid-Ajaj/Fable.Remoting/issues/134#issuecomment-1037897673 are required now instead), but the conversions are still there, just pushed further down the stack, so no performance is gained. To eliminate them completely, we'd first have to reshuffle Paket groups and TFMs, because the relevant Remoting projects either don't target `net6.0` or are pinned to FSharp.Core 4.7.2 owing to the `lowest_matching` constraint.

### DIFF
--- a/ClientV2Tests/src/App.fs
+++ b/ClientV2Tests/src/App.fs
@@ -1488,6 +1488,21 @@ let binaryServerTests =
                 test.equal output input
             }
 #endif
+
+        testCaseAsync "IBinaryServer.pureTask" <|
+            async {
+                let! output = binaryServer.pureTask
+                test.equal 42 output
+            }
+
+        testCaseAsync "IBinaryServer.echoMapTask" <|
+            async {
+                let input = ["hello", 1] |> Map.ofList
+                let! output = binaryServer.echoMapTask input
+                match input = output with
+                | true -> test.pass()
+                | false -> test.fail()
+            }
     ]
 
 let cookieServer =
--- a/ClientV2Tests/src/ClientV2.fsproj
+++ b/ClientV2Tests/src/ClientV2.fsproj
@@ -2,6 +2,7 @@
   <PropertyGroup>
     <TargetFramework>net6.0</TargetFramework>
     <LangVersion>latest</LangVersion>
+	<DefineConstants>TASK_AS_ASYNC</DefineConstants>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="..\..\Fable.Remoting.IntegrationTests\Shared\SharedTypes.fs" />
--- a/Fable.Remoting.IntegrationTests/Shared/ServerImpl.fs
+++ b/Fable.Remoting.IntegrationTests/Shared/ServerImpl.fs
@@ -2,6 +2,7 @@ module ServerImpl
 
 open SharedTypes
 open System
+open System.Threading.Tasks
 
 module Async =
     let result<'a> (x: 'a) : Async<'a> =
@@ -123,7 +124,9 @@ let serverBinary : IBinaryServer = {
     echoTimeOnlyMap = Async.result
     echoDateOnlyMap = Async.result
     echoRecordWithChar = Async.result
-}
+
+    pureTask = Task.FromResult 42
+    echoMapTask = fun map -> async { return map } |> Async.StartAsTask }
 
 // Async.result : 'a -> Async<'a>
 // a simple implementation, just return whatever value you get (echo the input)
--- a/Fable.Remoting.IntegrationTests/Shared/SharedTypes.fs
+++ b/Fable.Remoting.IntegrationTests/Shared/SharedTypes.fs
@@ -285,7 +285,7 @@ type IBinaryServer = {
     echoNestedAnonRecord : Maybe<{| nested: {| name: string |} |}> -> Async<Maybe<{| nested: {| name: string |} |}>>
 
     // mixed Task on the server, Async in JS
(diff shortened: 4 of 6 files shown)


Zaid-Ajaj:
Hi @kerams I understand this adds support for tasks on the server side of things but given that Fable does not support it, i think it would make sense to enable it on the _dotnet client_ too since that is the only place where it will be usable (I saw the type alias trick but I wouldn't recommend using it just to avoid an explicit `Async.AwaitTask` call)

kerams:
I can try, but it looks like it will involve a lot of (almost) duplicated code and many Task-Async to and fros, because of the reason I outlined above.

kerams:
I think my current proposed solution is a combination of 1 and 2. On the server side, the Remoting API would support `Task` returning functions, but internally there are still many unnecessary conversions, because I wanted to avoid duplicating too much library code and, more importantly, Remoting does not target a high enough version of FSharp.Core and TFMs. I'm sure this could be solved with a project, TFM and Paket group overhaul, but franky I don't know where to start with that. Perhaps the simplest way to proceed would be to do what Giraffe does? Release a new major version and make a clean break - from now on Remoting users on the server will need FSharp.Core 6 and target net6.0. Whoever has different requirements can still use previous versions. After all, the library has been stable for quite some time. Not sure what Zaid would think about that though. An alternative would be to maintain parallel packages/package versions (as is the case with the Fable.Remoting.GiraffeNET5 project), but that just seems to be a maintenance nightmare...

Without extra support for tasks in Fable, or coming up with some other clever way of sharing the API contract instead of a record with fields returning asyncs/tasks (a major breaking change of itself), option 3 is out of the question. We thus have to resort to type spoofing I mentioned earlier

```fsharp
#if !TASK_AS_ASYNC && !FABLE_COMPILER
type Async<'a> = System.Threading.Tasks.Task<'a>
#endif
```

The client Fable projects sees asyncs, so business as usual, but server developers can return native tasks to Remoting.

I'd love to hear ideas about something better, but as things stand, this might be the best we can do.

Thorium:
I tried this branch. Observations:

Server side works well:

- Indeed, this PR fixes the server runtime exception with tasks, so now the tasks come from server correctly and the method executes, the API works! Yay, when do you merge?

Client side I had issues:

- In the client side, there is no `Cmd.OfTask` because Fable-compiler doesn't have Async. In Elmish source there is [Cmd.OfTask](https://github.com/elmish/elmish/blob/d58b1514831ce3b3a78e9bfd7f555d516e960d87/src/cmd.fs#L203) but that is conditional-compilation not executed in Fable. 

- I cannot do `let cmd = Cmd.OfAsync.perform (myMethod >> Async.AwaitTask) () MyEvent` either because again there is no Async.AwaitTask in Fable, the website dies on runtime.

- The client fails to handle `Task`. There is `TypeInfo.Async` and `TypeInfo.Promise` handled in `Proxy.fs` and `Remoting.fs` but not Task. Should the `Task` be added to [Fable.SimpleJson TypeInfo](https://github.com/Zaid-Ajaj/Fable.SimpleJson/blob/73eb5aed779bc6c4a5efb90d796e87d96e7e5a29/src/TypeInfo.fs#L21) or should Fable understand to convert Tasks from API to Promise or something like that?

- From client side, it might be that I just didn't update some references properly, but as far as I can see, there might be some missing features here.

kerams:
@Thorium, the type spoofing only happens in the context of Fable, so it has no impact on server performance. Also, I assume the client doesn't work (based on your second comment) because you omitted the conditional compilation.

kerams:
> Typical webserver role is to fetch a resource from task-based asynchronous resource like "give me an item from database" or "read me a file". That's not an asynchronous workflow.

I don't get what you mean. `Async` and `Task` are conceptually the same thing, with slightly different technicalities (hot-cold, repeatability, performance).

> If you have 300 users online, it really starts to make a performance difference. Besides that, tasks have better stack-trace.

I haven't done any Task-Async benchmarks, but I can't imagine it would make a huge difference even when you have 300 users on your website concurrently. As noted here https://github.com/Zaid-Ajaj/Fable.Remoting/pull/293#issuecomment-1065863258, this PR only adds additional conversions when you use `Task` in your API, so the performance will actually be ever so slightly worse. I would like to remove all traces of `Async` from Remoting's internals, but it would involve bumping the FSharp.Core requirements (and maybe the minimum TFM to net6.0? I am not sure if it is required for task CEs).

Thorium:
Task is CLR optimized async small task.

Async is non-CLR-optimized old way to do complex asynchronous workflows. It works better on more complex scenarios, but it's poor on performance.

Library like Fable remoting is not all about asynchronous workflows, so it should migrate towards the new way as soon as possible.

If Fable is not going to support easy usage of task, meanehile all the other .NET (C# and now F#) libraries are taking the Task direction, will Fable's fate be exactly what happened to Suave: People try to find ways to migrate away from it. When the Suave maintainers finally understood, it was already too late.

kerams:
https://github.com/Zaid-Ajaj/Fable.Remoting/pull/323

kerams:
@Thorium, if it's performance you're after, I suggest you don't upgrade just yet. On my machine a single call is about 0.5ms slower than before (and this is tasks only, from top to bottom), possibly because of https://github.com/dotnet/fsharp/issues/12839#issuecomment-1264370200.

PR #294, opened 2022-02-20
More MsgPack perf

kerams:


Fable.Remoting.MsgPack/Write.fs
```diff
@@ -285,19 +308,23 @@ let inline writeUnion union (out: Stream) (caseSerializers: Action<'a, Stream>[]
...
-    // save 1 byte if the union case has a single parameter
-    if fieldSerializers.Length <> 1 then
-        // todo one byte less for no args too (change fixarr above)
-        writeArrayHeader fieldSerializers.Length out
+    if fieldSerializers.Length = 0 then
+        out.WriteByte (Format.fixarr 1uy)
+        out.WriteByte (Format.fixposnum tag)
+    else
+        out.WriteByte (Format.fixarr 2uy)
+        out.WriteByte (Format.fixposnum tag)
```
kerams, line 316: This will break the client if it's not updated at the same time as the server. Not sure if it deserves a major version bump or not (and not even that is going to prevent that scenario).

Zaid-Ajaj: We assume that every time someone is updating remoting packages, that they are updating both client and server to latest, so it's fine 🤞

Zaid-Ajaj:
Shall publish a batch or wait for #293 first (with dotnet client support)?

kerams:
Up to you, this is nothing critical.

Issue #301, opened 2022-05-26
Possibility of adding a global error handler on the web client?


enriquein:
Hello 😄. I was wondering if there could be a way to add an error handler to the web proxies just like we can have them in the server side. My specific use case for this would be that I want to perform a specific set of actions if any of the endpoints at any point return `401/UnAuthorized`. At the moment I'm using `React.useDeferred` and `React.useDeferredCallback` from Feliz, so I wrote a function that does the set of actions if the `ProxyRequestException` has a `StatusCode` of `401`. The issue is that in order for me to implement it, at the moment I need to place the function in each `match` block where I check the status of the deferred value. I worry that at some point someone in my team might forget to set the function properly, since we're used to using interceptors in other frameworks/libraries for these concerns.

Something like this maybe:
```
let musicStore = 
    Remoting.createApi()
    |> Remoting.withErrorHandler errorHandler  // to make the API similar/consistent with the server counterpart
    |> Remoting.buildproxy<IMusicStore>
```

enriquein:
Decided to take a stab at this and landed on what I think is a good happy medium. Instead of a global error handler, I made it possible to add a global interceptor. Hope this is useful to more people.

Zaid-Ajaj:
Hi there @enriquein, so this problem has popped before and in fact, we _did_ have a similar feature to what you implemented in #303 (it was years ago when we still used CEs for remoting builders), 

The reason we _removed_ was so that we do encourage users to handle errors globally. An error is ideally handled at the place where it happens. That said, it is sometimes useful to have a global error handler and the current API already allows for it. Let me explain, so instead of using an instance of a protocol API, let's call that `IServerApi`, you WRAP the function in another function that handles global errors and exceptions:

```fs
module Api

// creata an instance of the protocol record
let serverApi = 
    Remoting.createApi()
    |> Remoting.buildproxy<IServerApi>

// wrap it here to handle global errors
let call (f: IServerApi -> Async<'t>) = 
  async {
      try
        let! result = f serverApi
        return (Some result)
      with
      | ex -> 
          // Do something global
          // show toast
          // redirect to login
          return None
  }
``` 
Now instead of `Api.serverApi.funcName()` you would write `Api.serverApi.call (fun api -> api.funcName())` 

Yes, a bit more code but does what you need to do in this instance. Also you could make `call` return a union of things you care about, not just `Option<'t>`.

I hope this helps 🙏

enriquein:
[...]

PR #303, opened 2022-07-02
Add support for optionally adding an error interceptor. Addresses #301

enriquein:
This adds support  for optionally specifying an error interceptor for responses. The idea is that this interceptor would allow you to run centralized handlers for certain types of errors, without affecting the outcome of the response. For example, you could write an error interceptor to "handle" 401 responses by redirecting back to login, or a general 400-500 error interceptor that shows a toast notification. 

The signature of the interceptor is `statusCode: int -> responseBody: string -> unit`. You can set up the interceptor by calling `Remoting.withErrorInterceptor`: 

```
let errorInterceptor statusCode responseBody = 
    match statusCode with
    | 401 -> redirectToLogin()
    | 500 -> console.log "error"
    | _ -> () 

let musicStore = 
    Remoting.createApi()
    |> Remoting.withErrorInterceptor errorInterceptor  
    |> Remoting.buildproxy<IMusicStore>
```

### DIFF
--- a/Fable.Remoting.ClientV2/Proxy.fs
+++ b/Fable.Remoting.ClientV2/Proxy.fs
@@ -125,6 +125,11 @@ module Proxy =
                         let! responseText = Blob.readBlobAsText responseAsBlob
                         let response = { StatusCode = statusCode; ResponseBody = responseText }
                         let errorMsg = if n = 500 then sprintf "Internal server error (500) while making request to %s" url else sprintf "Http error (%d) while making request to %s" n url
+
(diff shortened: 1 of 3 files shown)


Zaid-Ajaj:
Hi there @enriquein, as mentioned in [my comment](https://github.com/Zaid-Ajaj/Fable.Remoting/issues/301#issuecomment-1204481586), the current API already allows for writing global interceptors when needed.

enriquein:
Closing this pull request since it's redundant. Thank you for the feedback!

Issue #304, opened 2022-07-06
Binary serialization bug with Result.Error


sWW26:
I'm getting some weird behaviour when using binary serialization with the `Result` type when the `Error` contains a string.

I've had a brief poke around to see if I can repro it using only the `Fable.Remoting.MsgPack` library directly however that seems to work fine.  It only seems to happen when going through the proxy / Giraffe?

I've put together a little sample app here that shows the issue: https://github.com/sWW26/FableBinarySerializationBug

The API always returns `Error "Test"` however the string ends up getting mangled.
![image](https://user-images.githubusercontent.com/6764717/177636003-a27cebcd-625a-4ad9-8065-574562bdaebf.png)

kerams:
Will take a look.

kerams:
Very strange - the first 2 responses contain corrupted (?) data. From the 3rd onwards, everything seems to be OK. Doesn't ring any bells, so I'll have to sleep on this.

kerams:
@sWW26, I've managed to ascertain this issue is deeper than Remoting. Could be due to JIT, but F# codegen is more likely. Follow the progress here https://github.com/dotnet/fsharp/issues/13447.

sWW26:
Thanks for investigating!

Zaid-Ajaj:
Closing as fixed since latest Fable.Remoting.MsgPack v1.18.0

PR #305, opened 2022-07-08
Work around stackalloc codegen bug

kerams:
https://github.com/Zaid-Ajaj/Fable.Remoting/issues/304, https://github.com/dotnet/fsharp/issues/13447

### DIFF
--- a/Fable.Remoting.MsgPack/Write.fs
+++ b/Fable.Remoting.MsgPack/Write.fs
@@ -228,6 +228,7 @@ let writeString (str: string) (out: Stream) =
 
         writeStringHeader bytesWritten out
         out.Write (Span.op_Implicit (buffer.Slice (0, bytesWritten)))
+        ()
     else
         let buffer = System.Buffers.ArrayPool.Shared.Rent maxLength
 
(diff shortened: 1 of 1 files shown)


sWW26:
Any chance this could get this merged & released?

kerams:
@Zaid-Ajaj, v1.17 with this fix sets an upper constraint on FSharp.Core for some reason. Any idea how this happened?

image

Zaid-Ajaj:
@kerams I think latest paket is more strict when you say `nuget FSharp.Core ~> 4.7.2` it means `>= 4.7.2 && < 4.8` but previously it didn't do that. Now I've changed the requirement to `nuget FSharp.Core >= 4.7.2` with strategy `lowest_matching: true` and published a new batch so it should be alright I think 🤞 

In any case, huge thanks for the fix 🙏  I would not have known how and where to start fixing this 😅 

@sWW26 A new batch of remoting packages have been published 🚀 , when upgrading, please make sure to update both client _and_ server when doing so.

sWW26:
That's great. Thanks both!

PR #313, opened 2022-09-17
Remove ITypeResolver to support Fable 4

kerams:
Support for type injection is removed from Fable(.Core) 4, hence this workaround.

I've thought about dropping the Fable 2 tests and adding 4, but versions 3 and 4 of the compiler are distributed as dotnet tools and I don't think it's possible to have multiple versions side-by-side.

### DIFF
--- a/Fable.Remoting.ClientV2/Remoting.fs
+++ b/Fable.Remoting.ClientV2/Remoting.fs
@@ -43,8 +43,8 @@ module Remoting =
         { options with CustomResponseSerialization = Some serializer }
 
 type Remoting() =
-    static member buildProxy<'t>(options: RemoteBuilderOptions, [<Inject>] ?resolver: ITypeResolver<'t>) : 't =
-        let resolvedType = resolver.Value.ResolveType()
+    /// For internal library use only.
+    static member buildProxy(options: RemoteBuilderOptions, resolvedType: Type) =
         let schemaType = createTypeInfo resolvedType
         match schemaType with
         | TypeInfo.Record getFields ->
@@ -98,3 +98,5 @@ type Remoting() =
         | _ ->
             failwithf "Cannot build proxy. Exepected type %s to be a valid protocol definition which is a record of functions" resolvedType.FullName
 
+    static member inline buildProxy<'t>(options: RemoteBuilderOptions) : 't =
+        Remoting.buildProxy(options, typeof<'t>)


Fable.Remoting.ClientV2/Remoting.fs
```diff
@@ -98,3 +98,5 @@ type Remoting() =
         | _ ->
             failwithf "Cannot build proxy. Exepected type %s to be a valid protocol definition which is a record of functions" resolvedType.FullName
 
+    static member inline buildProxy<'t>(options: RemoteBuilderOptions) : 't =
+        Remoting.buildProxy(options, typeof<'t>)
```
kerams, line 102: I've extracted the body of the function so that the rest of the code is not unnecessarily inlined and duplicated when using multiple remoting endpoints. Because of the inlining the extracted method has to be public too, but it should not cause any problems.

Fable.Remoting.ClientV2/Remoting.fs
```diff
@@ -43,8 +43,8 @@ module Remoting =
         { options with CustomResponseSerialization = Some serializer }
 
 type Remoting() =
-    static member buildProxy<'t>(options: RemoteBuilderOptions, [<Inject>] ?resolver: ITypeResolver<'t>) : 't =
-        let resolvedType = resolver.Value.ResolveType()
+    /// For internal library use only.
+    static member buildProxy(options: RemoteBuilderOptions, resolvedType: Type) =
```
Zaid-Ajaj, line 47: Can this be made `private`?

Fable.Remoting.ClientV2/Remoting.fs
Zaid-Ajaj: Looks good, though in this case I don't believe inlining has much benefit so we can remove the `inline` modifier here

Zaid-Ajaj (changes requested):
Changes look good, just commented some minor things. 

About testing both v3.x _and_ v4.x, I think we can add a build target that installs v4, runs the tests, then uninstalls it back to v3.x to make sure it still works but that can be added in another PR

Zaid-Ajaj:
OK you can ignore my comments, the inlining is still needed and I believe you cannot make the function body private when we do that.

kerams:
Yep, without `inline` Fable cannot insert the type object, and with `inline` the extracted method has to be public. I chose not to call it something like `buildProxyInternal`, so that it does not stand out in Intellisense. I thought it was a small price to pay for potentially reducing duplication in the bundle.

Zaid-Ajaj:
Published as of Fable.Remoting.Client v7.22

PR #314, opened 2022-09-17
Run integration tests using Fable v4 alongside v3

Zaid-Ajaj:
Basically
```fs
createTarget "IntegrationTestsNagareyama" <| fun _ ->
    runFableIntegrationTests()
    
createTarget "IntegrationTestsNagareyamaV4" <| fun _ ->
    withDotnetTool "fable" "4.0.0-theta-004" (fun _ -> runFableIntegrationTests())
```


kerams:
Clever. Fable.Core versions can potentially also cause problems, but I don't think it's something that we need to be concerned about in this case https://www.fuget.org/packages/Fable.Core/4.0.0-theta-001/lib/netstandard2.0/diff/3.7.1/

PR #323, opened 2022-10-01
Tasks

kerams:
Implements #134. Supersedes #293.

Bumped required FSharp.Core for the server bits to 6.0, so that the internals can be Task-based for extra performance. Just like in #293, Task endpoints are supported, but still cannot be consumed in Fable without the type spoof hack.

DotnetClient has been updated and supports Task endpoints. There are also additional methods on the Proxy type that return Task instead of Async: `let! result = proxy.callTask (fun server -> server.pureAsync)`.

The test failure comes down to the fact that 3 UI test projects reference FSharp.Core via the Main and Client dependency groups, and because of the lowest_matching constraint both version are different. Not sure how to solve this.


Zaid-Ajaj:
Hi @kerams sorry for taking this long to look into this PR, it looks great! I will merge and try to deal with the failing tests but all in all, we should get this out there 😄

Zaid-Ajaj:
Fixed the tests and published a batch of packages 🚀 should be up as soon as nuget finishes indexing

PR #325, opened 2022-10-18
Fix #324

### LINKED ISSUE #324: [BUG] 'withCredentials' is ignored – hence cross-site usage is broken
Hi

So far my experience with Fable.Remoting have been wonderful.

However today I ran into a bug. I have changed my hosting setup and now the frontend and backend are on two different domains. Hence I have to use CORS with credentials. But this seem to be broken i Fable.Remoting.Client.

It seems to be correct in 'Http.fs':
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Http.fs#L68

But in 'Proxy.fs' the option is simple ignored and never passed to 'Http':
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Proxy.fs#L113
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Proxy.fs#L117
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Proxy.fs#L143
https://github.com/Zaid-Ajaj/Fable.Remoting/blob/master/Fable.Remoting.Client/Proxy.fs#L147

All of these are:
```
|> Http.withHeaders headers
```

But needs to be:
```
|> Http.withHeaders headers
|> Http.withCredentials options.WithCredentials
```

I would consider this a pretty serious bug as it prevents usage of Fable.Remoting with a frontend on one domain and backend on another domain (assuming an authenticated use case where users are logged in – and hence cookies need to be sent).

I have tested my endpoint from dev tools on the frontend domain, and it works as expected when using:
```
const xhr = new XMLHttpRequest();
xhr.open('GET', '...endpoint-url...');
xhr.withCredentials = true;
xhr.send();
```

cr3wdayt5p:
Fixes #324

### DIFF
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -111,10 +111,12 @@ module Proxy =
                             Http.post url
                             |> Http.withBody requestBody
                             |> Http.withHeaders headers
+                            |> Http.withCredentials options.WithCredentials
                             |> Http.sendAndReadBinary
                         else
                             Http.get url
                             |> Http.withHeaders headers
+                            |> Http.withCredentials options.WithCredentials
                             |> Http.sendAndReadBinary
 
                     match statusCode with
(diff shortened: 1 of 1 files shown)


Zaid-Ajaj:
Thanks a lot for the fix @cr3wdayt5p, I've merged and published the fix as of v7.23 of Fable.Remoting.Client

PR #333, opened 2022-12-02
Accept Task-returning methods in JSON serialization

kerams:
This problem is hit only with JSON serialization when the contract file is in a separate project (which we don't use in tests here, so I can't add a case without creating extra projects - but I did try it though :)). 

@lukaszkrzywizna

### DIFF
--- a/Fable.Remoting.Client/Proxy.fs
+++ b/Fable.Remoting.Client/Proxy.fs
@@ -134,7 +134,13 @@ module Proxy =
                     match returnTypeAsync with
                     | TypeInfo.Async getAsyncTypeArgument -> getAsyncTypeArgument()
                     | TypeInfo.Promise getPromiseTypeArgument -> getPromiseTypeArgument()
-                    | _ -> failwithf "Expected field %s to have a return type of Async<'t>" func.FieldName
+                    | TypeInfo.Any getReturnType ->
+                        let t = getReturnType()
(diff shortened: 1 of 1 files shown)


Fable.Remoting.Client/Proxy.fs
```diff
@@ -134,7 +134,13 @@ module Proxy =
                     match returnTypeAsync with
                     | TypeInfo.Async getAsyncTypeArgument -> getAsyncTypeArgument()
                     | TypeInfo.Promise getPromiseTypeArgument -> getPromiseTypeArgument()
-                    | _ -> failwithf "Expected field %s to have a return type of Async<'t>" func.FieldName
+                    | TypeInfo.Any getReturnType ->
+                        let t = getReturnType()
+                        if t.FullName.StartsWith "System.Threading.Tasks.Task`1" then
+                            t.GetGenericArguments().[0] |> createTypeInfo
```
kerams, line 140: This could go into Fable.SimpleJson, but I assumed you wouldn't want it there as tasks aren't supported in Fable.

Zaid-Ajaj:
> This could go into Fable.SimpleJson (TypeInfo.Task), but I assumed you wouldn't want it there as tasks aren't supported in Fable.

Ideally, it should go in SimpleJson just like Async/Promise but that can be done later 😄 now merged and published as of Fable.Remoting.Client v7.25 🚀

lukaszkrzywizna:
It works like a charm! Thanks! <3

PR #345, opened 2023-07-16
Fix MsgPack on .NET 8

kerams:
Opening `System.Collections.Generic` on .NET 8, causes `AddRange` calls on a `ResizeArray` to be resolved to [an extension method taking a span](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.collectionextensions.addrange?view=net-8.0), instead of [an instance method on the list](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.addrange?view=net-8.0). Fable then complains that the implicit cast to `ReadOnlySpan` is not supported.

```
fable_modules\Fable.Remoting.MsgPack.1.18.0\Write.fs(575,25): (575,30) error FABLE: System.ReadOnlySpan`1.op_Implicit (static) is not supported by Fable
```

### DIFF
--- a/Fable.Remoting.MsgPack/Write.fs
+++ b/Fable.Remoting.MsgPack/Write.fs
@@ -2,7 +2,6 @@
 
 open System.IO
 open System
-open System.Collections.Generic
 open System.Text
 open FSharp.Reflection
 open FSharp.NativeInterop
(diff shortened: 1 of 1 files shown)


jwosty:
Just ran into this exact issue trying to upgrade my app to .NET 8. Would be awesome to have this merged!

Zaid-Ajaj:
Hi @kerams @jwosty this went a bit under my radar, apologies for that 🙏 I will try to merge and publish this upcoming weekend

jwosty:
No problem.

I also see that a lot of the CI tooling is out of date (it seems to build with .NET 3). Any way I can help upgrade some stuff?

Zaid-Ajaj:
Merged and published as of Fable.Remoting.MsgPack v1.19 🚀 apologies again for the delay

> Maybe I could start a .NET 8 branch, and have that use Fable 4 as the default (doing the same thing as https://github.com/Zaid-Ajaj/Fable.Remoting/pull/314, just with Fable 3 being the secondary build)? Then we can merge when .NET 8 comes out in October?

I feel like remoting doesn't have to be running on the bleeding edge of dotnet 8, though CI setup could use an upgrade

PR #357, opened 2023-12-30
Remoting.Client: dispose of the CancellationTokenRegistration when done

PierreYvesR:
Otherwise the XMLHttpRequest is not being garbage-collected (observed in Firefox, Chrome, and Edge)

### DIFF
--- a/Fable.Remoting.Client/Http.fs
+++ b/Fable.Remoting.Client/Http.fs
@@ -56,10 +56,11 @@ module Http =
             | Some f ->  f xhr
             | _ -> ignore()
 
-            token.Register(fun _ ->
-                xhr.abort()
-                cancel(System.OperationCanceledException(token))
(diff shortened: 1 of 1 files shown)


kerams:
Interesting. So this causes a memory leak with every request in every application?

PierreYvesR:
When using the Fable (javascript) client library, it seems so.
I've used a freshly-created SAFE-stack template for testing (as per https://safe-stack.github.io/docs/quickstart/): the number of XMLHttpRequest never goes down in the memory profiler.
On an app doing a few 1MB requests per minute, memory would run out after a few hours at around a few GB.

Zaid-Ajaj:
Thanks @PierreYvesR I've merged and published this as of Fable.Remoting.Client v7.30 please take it for a spin and let us know if everything is alright

PR #365, opened 2024-05-01
Adds XML docs to the `RouteInfo` type.

ChrSteinert:
I just happen to pass those by. When I was looking at that type I felt some (even if small) annotations missing, so I added them.


Zaid-Ajaj:
Merged and published as of latest Fable.Remoting.Server v5.38.0 🚀

PR #366, opened 2024-07-08
Minor code clean-up: Avoid dictionary double-lookups.

Thorium:
Avoid dictionary double-lookups.


Zaid-Ajaj:
Thank you @Thorium 🙏 merged and published as of Fable.Remoting.Json v2.24.0 🚀

PR #370, opened 2024-09-22
Update mime type for MsgPack

kerams:
[An official type has been assigned](https://www.iana.org/assignments/media-types/application/vnd.msgpack), so let's use that.


Zaid-Ajaj:
merged and published 🚀

PR #372, opened 2024-12-31
Update to .NET 8

DragonSA:
With .NET 6 [EOS](https://devblogs.microsoft.com/dotnet/dotnet-6-end-of-support/) I've updated all targets to .NET 8 (and removed redundant targets).

Sponsored by [Counterpoint Dynamics](https://cpdynamics.co.za)


Zaid-Ajaj:
Hi there @DragonSA 

I really appreciate the support and the contribution! Can you please share the motivation behind this PR? Are you unable to use the Remoting packages because they are targeting .NET 6?

DragonSA:
Hi

It's a pleasure.  The main reason is the recent update to Rider broke loading of projects without a `global.json` targeting a specific .NET version.  So, I chose the current LTS.  

This enabled me to work on a feature I would like to propose: support for Blazor applications.  This would involve:
- Support for Func<>
- Support for POCO (and C# records)
- Extension methods that work with C#, ASP.NET and dependency injection.  

I have the above almost to the point of a work proof of concept.  I was planning to then share and propose the changes.  But, since we are chatting now I thought why not get your initial thoughts on it?

Zaid-Ajaj:
> The main reason is the recent update to Rider broke loading of projects without a global.json targeting a specific .NET version. So, I chose the current LTS.

@DragonSA I would rather not update unless necessary or bringing better performance for the libraries because otherwise it's just more maintenance

>  This enabled me to work on a feature I would like to propose: support for Blazor applications. 

Honestly I don't understand how Blazor fits in here 🤔 can you elaborate on your use case? If it a scenario of using Fable.Remoting on the backend and Blazor on the frontend, shouldn't that already be possible using `Fable.Remoting.DotnetClient` from the Blazor frontend? Define your F# types and share those in a class library with the Blazor frontend, then `Fable.Remoting.DotnetClient` should be able to handle these types. 

That said, I am not a Blazor expert so I don't know what could be blocking here

Zaid-Ajaj:
Implemented in #391

PR #380, opened 2025-05-20
Fix: Turkish locale breaks discriminated union deserialization

aaronmu:
Adds a test case that reproduces the issue with discriminated union deserialization when the current culture is set to Turkish (tr-TR). The test demonstrates how culture-sensitive string comparisons in TypeShape can cause union case name matching to fail.

The test sets the thread's CurrentCulture and CurrentUICulture to Turkish before running the test cases and restores the original culture afterward.

### DIFF
--- a/Fable.Remoting.MsgPack.Tests/FableConverterTests.fs
+++ b/Fable.Remoting.MsgPack.Tests/FableConverterTests.fs
@@ -13,6 +13,9 @@ let equal x y = Expect.equal true (x = y) (sprintf "%A = %A" x y)
 let pass() = Expect.equal true true ""
 let fail () = Expect.equal false true ""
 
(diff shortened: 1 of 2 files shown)


aaronmu:
Special thanks to my good friend and colleague @jonasdw who found the problem and actually also came up with the fix but now is a bit salty because he was too late to contribute. Couldn't have done it without you mate!

kerams (approved):

aaronmu:
@Zaid-Ajaj @kerams , any idea when this could be merged?

Zaid-Ajaj:
I can merge and publish this week

Zaid-Ajaj:
Latest batch should include this in Fable.Remoting.MsgPack and all dependant packages

